PR #8, opened 2015-03-04
switch from ruby sass to node sass

xiwcx:
- add npm package, remove a bunch of ruby gems
- remove sass-globbing, replace with direct imports
- remove mixins that rely on `@at-root`


chuckcarpenter:
Still requires Ruby? Also, if this code is a major bump, I'd make the changes to the JSON version and we can tag on merge.

xiwcx:
ruby is needed for hologram only to compile the styleguide. i'm gonna wait to tag a release as i have a feeling a lot of new stuff is coming down the pipeline.

PR #14, opened 2015-07-14
SYTO-234: Move to mortar styles added recently for onepager in MC's Alice app

davidocoulter:
https://jira.nationalgeographic.com/browse/SYTO-234

This PR adds styles intended to be folded back into Mortar, although we may need to reevaluate naming and make some things more generic.

[ADDED] Styles from MC, which may or may not belong in Mortar

To test: 
- Run `grunt serve`
- Ensure nothing breaks
- :balloon:

### DIFF
--- a/app/src/styles/_shame.scss
+++ b/app/src/styles/_shame.scss
@@ -1,2 +1,148 @@
 // Things with no home
 // ===================
+
+
+// Styles from Member Center
+
+// typography
+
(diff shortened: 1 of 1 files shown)


chuckcarpenter:
I think the intent is that this would need to be an accepted contribution, so renaming should be part of it and not just dumping in 'shame' file.

PR #17, opened 2015-12-01
[AEM-3353] Remove text selection color

bpearsoncse:
https://jira.nationalgeographic.com/browse/AEM-3353

Hey I'm Ben Pearson the new Front End Dev @CSE. We have a recommendation from @asabaylus to change this at the Mortar level. Let us know you would rather we override it with AEM.

### DIFF
--- a/app/src/styles/_base.scss
+++ b/app/src/styles/_base.scss
@@ -19,17 +19,15 @@ html {
     box-sizing: inherit;
 }
 
-// Text highlighting color
+// Use default selection color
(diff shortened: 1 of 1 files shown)


chuckcarpenter:
@xiwcx do we know what the base requirement is? @ellenbutters we used to have a default selection color. Should we continue that and have @bpearsoncse override or change the default?

xiwcx:
I thought it was a neat detail, but I'm not wedded to it.

PR #22, opened 2016-02-23
Feature/mortar 31

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-31


vamidjo:
left .mt_textarea in there since textarea has a specified min-height. 
Didn't include widths for input fields because they might conflict with other input types down the line. Don't mind including them however, if necessary.

PR #26, opened 2016-02-29
Mortar-30 Buttons

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-30

Testing instructions: 

1. git checkout
2. gulp serve
3. Visit http://localhost:3000/patterns

### DIFF
--- a/app/site/patterns/index.html
+++ b/app/site/patterns/index.html
@@ -4,7 +4,8 @@ title: UI Patterns
 
 <h3> Buttons </h3>
 
-<div>
+<div style="padding: 1em;">
+
   <button class="mt_btn mt_btn-default"> Default </button>
   <button class="mt_btn mt_btn-secondary"> Secondary </button>
   <button class="mt_btn mt_btn-naked"> <span> Naked </span> </button>
@@ -13,13 +14,10 @@ title: UI Patterns
   <button class="mt_btn mt_btn-default--inactive"> Default Inactive </button>
   <button class="mt_btn mt_btn-secondary--inactive"> Secondary Inactive </button>
   <button class="mt_btn mt_btn-naked mt_btn-naked--inactive"> Naked Inactive </button>
+
 </div>
 
-<br /> <br />
-
-<div class="mt_bg-color-neutral--darker">
-
-  <br /> <br />
+<div style="padding: 1em;" class="mt_bg-color-neutral--darker">
 
   <button class="mt_btn mt_btn-default"> Default </button>
   <button class="mt_btn mt_btn-secondary--reversed"> Secondary Reversed </button>
@@ -30,13 +28,9 @@ title: UI Patterns
   <button class="mt_btn mt_btn-secondary--inactive"> Secondary Inactive </button>
   <button class="mt_btn mt_btn-naked mt_btn-naked--inactive"> Naked Inactive </button>
 
-  <br /> <br />
-
 </div>
 
-<div class="mt_bg-color-neutral--dark">
-
-  <br /> <br />
+<div style="padding: 1em;" class="mt_bg-color-neutral--dark">
 
   <button class="mt_btn mt_btn-default"> Default </button>
   <button class="mt_btn mt_btn-secondary--reversed"> Secondary Reversed </button>
@@ -46,7 +40,5 @@ title: UI Patterns
   <button class="mt_btn mt_btn-default--inactive"> Default Inactive </button>
   <button class="mt_btn mt_btn-secondary--inactive"> Secondary Inactive </button>
   <button class="mt_btn mt_btn-naked mt_btn-naked--inactive"> Naked Inactive </button>
-
-  <br /> <br />
 
 </div>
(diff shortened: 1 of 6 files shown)


app/site/patterns/index.html
```diff
@@ -0,0 +1,52 @@
...
+  <button class="mt_btn mt_btn-secondary"> Secondary </button>
+  <button class="mt_btn mt_btn-naked"> <span> Naked </span> </button>
+  <button class="mt_btn mt_btn--success"> Success </button>
+  <button class="mt_btn mt_btn--error"> Error </button>
+  <button class="mt_btn mt_btn-default--inactive"> Default Inactive </button>
+  <button class="mt_btn mt_btn-secondary--inactive"> Secondary Inactive </button>
+  <button class="mt_btn mt_btn-naked mt_btn-naked--inactive"> Naked Inactive </button>
+</div>
+
+<br /> <br />
```
xiwcx: please use the gutter classes from the layout system so we can avoid presentational HTML. `.mt_row-gut-2` should do the trick here.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
+$mt_btn-inactive-color: rgba(128, 128, 128, 0.4);
```
xiwcx: please avoid one off colors, just use the closest shade of grey that we have in the palette already and we'll adjust as needed.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+
+button,
```
xiwcx: for now, let's avoid resetting any element selectors.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+&.mt_btn {
```
xiwcx: this `&` won't actually be do anything. since this selector is at the root, it doesn't have a parent to reference.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+  @extend .mt_h5;
```
xiwcx: rather than extending the class directly let's create a placeholder class like so:

```scss
%mt_h5 {
  // styles
}
```

then both the type class and this button class can reference that placeholder class instead.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+  border: 1px solid transparent;
+  border-radius: 0;
+  cursor: pointer;
+  text-align: center;
+  transition: 400ms;
+  background-color: transparent;
+
+  &.mt_btn-default{
+    background-color: $color-primary;
+    &:hover{
```
xiwcx: please include a hard return before every new "block" like you have on line 21 of this file

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+&.mt_btn {
+  @extend .mt_h5;
+  padding: 16px 32px 16px 32px;
+  margin: 0 10px 10px 0;
+  border: 1px solid transparent;
+  border-radius: 0;
+  cursor: pointer;
+  text-align: center;
+  transition: 400ms;
+  background-color: transparent;
```
xiwcx: make sure properties are alphabetized within blocks.

app/styles/patterns/_buttons.scss
```diff
@@ -0,0 +1,94 @@
...
+    }
+    &--inactive{
+      background: $color-neutral;
+      color: $mt_color-neutral-darkest;
+      cursor: default;
+      opacity: 0.3;
+    }
+  }
+
+  &.mt_btn-secondary{
```
xiwcx: let's break out secondary and naked as their own styles so they don't have to be chained together.

app/styles/layout/columns.html
```diff
@@ -4,7 +4,7 @@
 
 <h5>Available classes</h5>
 
-<p>"`*`" can be replaced with a relevant number. Column, offset, push, and pull are all respective to the twelve column grid so 1 through 12 are available. Gutters are avaiable in half, 1, 2, 3, and 4.</p>
+<p>"`*`" can be replaced with a relevant number. Column, offset, push, and pull are all respective to the twelve column grid so 1 through 12 are available. Gutters are available in half, 1, 2, 3, and 4.</p>
```
xiwcx: haha, nice catch!

app/styles/theme/_fonts.scss
```diff
@@ -33,9 +43,7 @@
 }
 
 .mt_h5 {
-  font-family: 'Pragmatica-web';
-  font-size: 14px;
-  font-weight: 600;
+  @extend %mt_h5;
```
xiwcx: would you remind repeating this process for each of the typographic styles and we'll call it a day on this ticket?

vamidjo: sure!

PR #28, opened 2016-03-02
change over to relative sizes, swith px values to variables

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-47

- verify that fonts have remained the same size after px -> em switch
- verify that where possible, px values have been changed to em


vamidjo:
ok looks good

PR #29, opened 2016-03-02
Feature/mortar 53

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-53

Testing instructions: 

1. git checkout
2. gulp test
3. Look for '0 errors'

### DIFF
--- a/app/styles/theme/_fonts.scss
+++ b/app/styles/theme/_fonts.scss
@@ -1,6 +1,6 @@
 // Extends
 
-%mt-h1 {
+%mt_h1 {
   font-family: 'Chronicle Display 3r';
   font-size: 65px;
   font-weight: 300;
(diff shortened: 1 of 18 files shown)


app/styles/theme/_fonts.scss
```diff
@@ -1,6 +1,6 @@
 // Extends
 
-%mt_h1 {
+%mt-h1 {
```
xiwcx: these need to be put back and the rule needs to be changed. that `mt_` preface should be on everything:

```yaml
rules:
  ...
  placeholder-name-format:
    - 2
    -
      convention: 'mt_([a-z0-9\-]+)'
      convention-explanation: 'All placeholder names must be prefaced with "mt_" and contain only lowercase letters and dashes afterward'
```

https://github.com/sasstools/sass-lint/blob/develop/docs/rules/placeholder-name-format.md

PR #35, opened 2016-03-10
Feature/mortar 55

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-55

Testing Instructions: 

1. Git checkout feature/mortar-55
2. Gulp serve
3. Visit http://localhost:3000/theme/


app/styles/theme/fonts.html
```diff
@@ -1,3 +1,5 @@
+<!-- The following is for review purposes. Remove this block once everything checks out -->
+
 <h3>Typography</h3>
 
 <div class="text-examples">
```
vamidjo: Left this file in here for review purposes. I can go ahead and remove it once everything checks out. @xiwcx

app/styles/_shame.scss
```diff
@@ -78,6 +78,24 @@
 }
 
 // ====================
+// ====================
+// Typography page
+// ====================
+// ====================
+
+.typographyexamples {
```
xiwcx: not a note yet, but if we find that this pattern is used elsewhere, we need to make sure to abstract it in the future.

vamidjo: Agreed.

PR #46, opened 2016-03-16
Feature/mortar 59

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-59

- git checkout 
- gulp serve
- Visit http://localhost:3000/home


vamidjo:
Note:
- Internal links are missing 
- Having trouble with the thumbnails

app/styles/home/feedback.html
```diff
@@ -0,0 +1,11 @@
+<article class="pagewrapper mt_col-gut-half">
```
xiwcx: Throughout this pull request I don't think the spec for [article](http://html5doctor.com/element-index/#article) or section [http://html5doctor.com/element-index/#section] are satisfied. In this and a few other instances the markup could be simplified:

```html
<div class="pagewrapper mt_row mt_row-gut-2">
    <div class="mt_col-lg-8 mt_col-sm-12 mt_col-gut-half">
```

app/styles/home/feedback.html
```diff
@@ -0,0 +1,11 @@
...
+  <section class="mt_row mt_row-gut-2">
+      <div class="mt_col-lg-8 mt_col-sm-12">
+        <div class="mt_h3">Send Us Your Feedback</div>
```
xiwcx: However, this header should be a header! Part of the reason the header type classes are abstracted is so that they can be applied to any heading type. For example:

```html
<h2 class="mt_h3">Send Us Your Feedback</h2>
```

app/styles/home/documentation.html
```diff
@@ -0,0 +1,141 @@
+<article class="pagewrapper mt_col-gut-half">
```
xiwcx: this is the exception, this could probably be considered an article. a section within an article is probably redundant though. I'm sorry to be so nitpicky, but since developers are going to be doing a lot of cut and paste from this site for example all the markup should be 💯

app/styles/home/feedback.html
vamidjo: A div is being used here instead of a heading because the h1-h4 come with their browser-added margins causing the spacing between elements to be different from the designs. How do we want to handle this? Should we special case individual headings and apply a specified margin whenever possible? or add a mortar classname (.mt_stripmargin) which would take care of that?

PR #58, opened 2016-04-04
Feature/mortar 67

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-67

Test instructions: 

- Git checkout
- Gulp serve
- Visit http://localhost:3000/patterns


app/modules/modals/_styles.scss
```diff
@@ -17,6 +17,7 @@
   top: 7.5vh;
   transition: .2s;
   width: 85vw;
+  z-index: 999;
```
xiwcx: z values should all be variables. make a new file: `global/vars/_z.scss`

it should contain

```scss
$z-bottom: -1;
$z-middle: 0;
$z-top: 999;
```

then reference that variable.

app/styles/_shame.scss
```diff
@@ -238,10 +239,10 @@ td, th {
 }
 
 .dashed-border {
-  background-image: -webkit-linear-gradient(top,$mt_color-primary 50%,transparent 0);
-  background-image: linear-gradient(to bottom,$mt_color-primary 50%,transparent 0);
+  background-image: -webkit-linear-gradient(top, $mt_color-primary 50%,transparent 0);
```
xiwcx: did autoprefixer miss this? if not, you shouldn't need the vendor prefixes.

app/styles/blocks/_counters.scss
```diff
@@ -14,13 +14,20 @@
 }
 
 .dotcounter-container {
```
xiwcx: missint `.mt_`

app/styles/blocks/_videopromo.scss
```diff
@@ -10,15 +10,15 @@
   background-size: contain;
   background-image: $mt_video-promo-play-button;
   border: none;
+  content: $mt_unit;
```
xiwcx: what's going on here?

app/styles/blocks/_counters.scss
vamidjo: does this go for `.dotcounter-item` as well?

xiwcx: goes for any and every class at the mortar level, but git blame says it was me who added all these!! my bad...

xiwcx:
screen shot 2016-04-07 at 10 24 09 am

Looks like there are some color classes that need to get updated!

PR #63, opened 2016-04-06
add gulp production task

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-99

- run `npm install -g http-server` or use whatever you like that can serve static files from a single directory
- `npm i`
- test `gulp build` against `gulp build --production`. make sure both work.
- after running `gulp build --production` cd to `.tmp/site` from the root of the mortar repository. run `http-server` (or comparable task) to check that the production build works (check every page)


asabaylus:
@xiwcx very nice works swell. One comment ... its not clear where production artifact is stored

PR #74, opened 2016-04-18
Feature/mortar 103

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-103

git checkout
gulp serve
Visit http://localhost:3000/typography

### DIFF
--- a/app/styles/theme/_pullquotes.scss
+++ b/app/styles/theme/_pullquotes.scss
@@ -1,6 +1,18 @@
 %mt_pullquote {
   p {
     margin-top: 0;
+
+    &:first-child::before {
+      content: '\201C';
+
+      @media #{$mt_mq-lg} {
+        margin-left: -$mt_fs-sm;
+      }
+    }
+
+    &:last-of-type::after {
+      content: '\201D';
+    }
   }
 
   border-bottom: 1px solid $mt_color-neutral-xl;
@@ -16,10 +28,4 @@
 
 .mt_pullquote {
   @extend %mt_pullquote;
-}
-
-.mt_pullquote-q {
-  @media #{$mt_mq-lg} {
-    margin-left: -$mt_fs-sm;
-  }
 }
--- a/app/styles/theme/_pullquotes.scss
+++ b/app/styles/theme/_pullquotes.scss
@@ -2,7 +2,7 @@
   p {
     margin-top: 0;
 
-    &:first-child::before {
+    &:first-of-type::before {
       content: '\201C';
 
       @media #{$mt_mq-lg} {
--- a/app/styles/theme/_pullquotes.scss
+++ b/app/styles/theme/_pullquotes.scss
@@ -1,0 +1,33 @@
+%mt_pullquote {
+  p {
+    @extend %mt_h3;
+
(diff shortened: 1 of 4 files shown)


app/styles/theme/_pullquotes.scss
```diff
@@ -0,0 +1,36 @@
+%mt_pullquote {
+  p {
```
xiwcx: screen shot 2016-04-19 at 10 17 41 am

this breaks down when there's more than one paragraph. probably you can just dispense with the nested style. the only thing that might be necessary is defining the distance between paragraphs.

app/styles/theme/_pullquotes.scss
```diff
@@ -0,0 +1,36 @@
...
+  @media #{$mt_mq-lg} {
+    padding: $mt_fs-xxxlrg / 2;
+  }
+}
+
+.mt_pullquote {
+  @extend %mt_pullquote;
+}
+
+.mt_pullquote--footer {
```
xiwcx: we can hold off on this until it's necessary, at the moment it just explicitly states a thing we can reasonably assume.

app/styles/theme/_pullquotes.scss
xiwcx: screen shot 2016-04-21 at 10 29 05 pm

i'm still seeing the same issue; quotations between every paragraph.

app/styles/theme/_pullquotes.scss
```diff
@@ -0,0 +1,25 @@
...
+  @media #{$mt_mq-lg} {
+    padding: $mt_fs-xxxlrg / 2;
+  }
+}
+
+.mt_pullquote {
+  @extend %mt_pullquote;
+}
+
+.mt_pullquote-q {
```
xiwcx: what was wrong with the ::before psuedo class? I'd prefer to avoid an extra element if possible.

app/styles/theme/_pullquotes.scss
vamidjo: Are you duplicating this? 

`<p class="mt_h3">
<span class="mt_pullquote-q">&ldquo;</span>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus magna. Cras in mi at felis aliquet congue.&rdquo;</p>`

There's a span with a classname in there. You'd have to exclude it. The quotations are inserted manually since using a :before pseudo class would insert them each time there's a paragraph. Is there a better way to do this?

Also, the span has a classname because the 'quotation mark' position isn't  consistent across desktop and mobile.

https://jira.nationalgeographic.com/browse/MORTAR-103

app/styles/theme/_pullquotes.scss
vamidjo: See above ^

app/styles/theme/_pullquotes.scss
xiwcx: Ah, I see now! Yes, there is a better way to do this, you can combine psuedoclasses:

```scss
.mt_pullquote {
  p:first-child::before {
    // styles here
  }
}
```

PR #124, opened 2016-06-03
Make snippets wrap on mobile

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-191

- git checkout
- gulp serve
- Verify that code snippets wrap on mobile

### DIFF
--- a/app/site/_styles/docs.scss
+++ b/app/site/_styles/docs.scss
@@ -412,3 +412,14 @@ pre {
 code {
   font-size: 14px;
 }
+
+pre,
+code {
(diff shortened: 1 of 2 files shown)


app/site/_styles/docs.scss
```diff
@@ -416,3 +416,14 @@ pre {
 code {
   font-size: 14px;
 }
+
+pre,
+code {
+  &[class*="language"] {
```
xiwcx: just group this with the rest of the styles so there's not a redundant block then i'll merge.

PR #132, opened 2016-06-10
add react-slick slider to MT slider

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-182

* `gulp serve`
* view code to make sure the react-slide component is being properly included and used
* view documentation (http://localhost:3000/components/) to ensure the content fromt he markup is being passed to the slider

### DIFF
--- a/app/modules/slider/docs.html
+++ b/app/modules/slider/docs.html
@@ -5,8 +5,13 @@
       data-pestle-module="Slider"
       data-pestle-options='{
         "transitionSpeed": 300,
-        "transitionType": "slide",
-        "initialSlide": 2
-      }'></div>
+        "transitionType": "fade",
+        "initialSlide": 2,
+        "slides": [
+          "http://www.fillmurray.com/g/600/400",
+          "http://www.placecage.com/600/400"
+        ]
+      }'
+      ></div>
   </div>
 </div>
--- a/app/modules/slider/docs.html
+++ b/app/modules/slider/docs.html
@@ -8,8 +8,14 @@
         "transitionType": "fade",
         "initialSlide": 2,
         "slides": [
-          "http://www.fillmurray.com/g/600/400",
-          "http://www.placecage.com/600/400"
+          {
+            "type": "image",
+            "src": "http://www.fillmurray.com/g/600/400"
+          },
+          {
+            "type": "image",
+            "src": "http://www.placecage.com/600/400"
+          }
         ]
       }'
       ></div>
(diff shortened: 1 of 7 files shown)


krlospelaez:
It looks pretty good for me. 

Maybe do you want to consider a "slide markup" agnostic approach? I think we can reuse it, not just for images but also for videos or more complex html structures.

The code would be something like this https://gist.github.com/krlospelaez/1a4cfedabf37d38d7387384be91234d7

@xiwcx @asabaylus What do you think about this approach?

xiwcx:
@krlospelaez, i'm a little lost in your example, not sure how this would help make the children more generic as the example still looks specific to. what would the markup look like? my suggestion would be more like:

```html
<div
data-pestle-module="Slider"
data-pestle-options='{
  "transitionSpeed": 300,
  "transitionType": "fade",
  "initialSlide": 2,
  "slides": [
    {
      "type": "image"
      "src": // single url or arrary or urls and srcsets      
    },
    {
      "type": "markup"
      "content": "<div>thing</div>"      
    }
  ]
}' ></div>
```
This way the slider component knows whether to pass the child to the picture component, render the markup straight up, or pass it to some other component like video. Is this too much?

krlospelaez:
Oh, I understand your confusion. The generic part is the react component (MTSlider). For example you can use existing MTSlider and create a PhotoGallery component:

```javascript
import React, {Component} from 'react';
import MTSlider from 'MTSlider.jsx';

export default class MTPhotoGallery extends Component {
   render() {
       return (
           <MTSlider>
               <div className="description">Promo Description</div>
               <img src="promo.jpg" alt="" />
           </MTSlider>
       );
   }
}
```

```javascript
import ReactDOM from 'react-dom';
import {Module} from 'pestle';

import MTPhotoGallery from 'MTPhotoGallery';

class MTPhotoGalleryPestle extends Module {
    init() {
       ReactDOM.render(<MTPhotoGallery />, this.el);
    }
}
...
```

So as you can see in the example above we're reusing the MTSlider component and creating a new one.

Now returning to this ticket, you should see MTSliderPestle.js just like documentation module to explain people how to use MTSlider but no as final component. So probably on this case a Pestle component is not mandatory because the MTSlider requires a useful markup like the html markup for a PhotoGallery.

I think on the incoming tickets we're going to create the PhotoGallery so we'll need to create a MTPhotoGallery react component with its own html markup and it'll reuse the MTSlider you're creating here. This PhotoGallery is going to have a MTPhotoGalleryPestle module to include the MTPhotoGallery on the page. This MTPhotoGalleryPestle is mandatory because MTPhotoGallery is a final component, where MTSlider isn't.

I saw your suggestion about the markup but remember we're using react for managing all html markup.

krlospelaez:
LGTM 👍

PR #133, opened 2016-06-13
Feature/mortar 173

vamidjo:
https://jira.nationalgeographic.com/browse/MORTAR-173

git checkout
gulp serve
Make sure all `.mt_cf` classes are replaced with .`mt_row`


xiwcx:
my only note is that the rows partial should still live in the layout directory and the cf partial should stay deleted.

PR #142, opened 2016-06-17
Feature/mortar 37

hkekeli:
https://jira.nationalgeographic.com/browse/MORTAR-37

git checkout
gulp serve
http://localhost:3000/patterns 
Test select menu. First option should be hidden and .mt_neutral-color-xd should remain when you click outside the select menu

### DIFF
--- a/app/styles/patterns/inputs.html
+++ b/app/styles/patterns/inputs.html
@@ -43,11 +43,12 @@
         <textarea class="mt_input mt_textarea mt_fullwidth mt_subh4 mt_row-gut-1" placeholder="I am a text area"></textarea>
 
         <!-- Select menu -->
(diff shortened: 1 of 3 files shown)


app/styles/patterns/inputs.html
```diff
@@ -44,10 +44,11 @@ <h3 class="mt_h3">Form Inputs</h3>
 
         <!-- Select menu -->
         <div class="mt_selectwrapper mt_selectwrapper--fullwidth">
```
xiwcx: please make sure that as you delete classes from the stylesheets you also delete them from the reference. this is the kind of thing that will send future developers on a wild goose chase.

hkekeli: I did remove it. Not sure what's happening but I keep having to make this update and alphabetize declarations on _select.scss repeatedly.

PR #143, opened 2016-06-20
Create mt_intratio--circle class

hkekeli:
https://jira.nationalgeographic.com/browse/MORTAR-206

git checkout
gulp serve
`.mt_round` is removed from code and replaced by `.mt_intratio--circle`

### DIFF
--- a/app/styles/utils/_circle.scss
+++ b/app/styles/utils/_circle.scss
@@ -1,0 +1,5 @@
+.mt_intratio--circle {
+  @extend .mt_intratio--square;
+
+  border-radius: 50%;
+}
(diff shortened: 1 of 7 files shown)


app/styles/utils/_circle.scss
```diff
@@ -1,3 +1,3 @@
-.mt_round {
+.mt_intratio--circle {
   border-radius: 50%;
```
xiwcx: i believe the point is that this is supposed to maintain the 1:1 aspect ratio without adding the square class. you can add this:

```scss
.mt_intratio--circle {
  @extend .mt_intratio_square;

  border-radius: 50%;	
}
```

PR #144, opened 2016-06-20
add custom theme to slider

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-180

1. view at [http://localhost:3000/components]()
- run `gulp serve`
- test on several browsers and devices using http://[your LAN ip].xip.io:3000/components
- test element queries (buttons should be different sizes depending upon how large the slider is rendered)
- test buttons
- test for swipe


krlospelaez:
LGTM 👍

PR #146, opened 2016-06-22
add emitted event to slide change

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-189

- `gulp serve`
- use arrow or keyboard to navigate slide change event
- either add a log or breakpoint to see the data payload


app/modules/slider/scripts/MTSlider.jsx
```diff
@@ -35,8 +36,18 @@ class MTSlider extends Component {
     }
   }
 
+  emitEvent(data) {
+    const slideData = {
+      currentSlideIndex: data,
+      thisSlide: this
+    }
+
+    PubSub.publish('MTSlider', slideData);
```
krlospelaez: @xiwcx Can you use this convention https://github.com/natgeo/javascript#events-objectname? Create a new file:

```javascript
export default const events = {
    slideChange: 'MTSlider.slideChange'
};
```
Then use it:

```javascript
import events from './events';
...
PubSub.publish(events.slideChange, slideData);
...
```

xiwcx: yeah, of course! thanks for the tip.

app/modules/slider/scripts/MTSlider.jsx
```diff
@@ -35,8 +36,18 @@ class MTSlider extends Component {
     }
   }
 
+  emitEvent(data) {
+    const slideData = {
+      currentSlideIndex: data,
+      thisSlide: this
```
krlospelaez: I think we don't need to pass thisSlide object, we can use pub/sub to communicate back if needed.

app/modules/slider/scripts/MTSlider.jsx
```diff
@@ -35,8 +36,18 @@ class MTSlider extends Component {
     }
   }
 
+  emitEvent(data) {
```
krlospelaez: onChangeSlide probably is a meaningful name?

krlospelaez:
Some unit tests to check the event is working would be great

app/modules/slider/scripts/MTSlider.jsx
xiwcx: the reason for "thisSlide" is that without it we don't actually know which gallery emitted the event which means we don't know, for example, which ad to rotate. is there another way?

krlospelaez: Ok, on this case you should give it a name or id to identify it but always using pub/sub to communicate both components. This will ensure loose coupling.

krlospelaez: When rendering MTSlider you should do something like this:

```javascript
<MTSlider identifier="xxxxxx" />
```

Inside MTSlider definition:
```javascript
...
onSlideChange(currentSlide) {
    const slideData = {
        identifier: this.props.identifier,
        currentSlide
    }
    PubSub.publish(events.slideChange, slideData);
}
....
```
So now you know which ad rotate based on identifier :)

xiwcx:
i may need some help writing these tests, not sure how to test the pubsub.

krlospelaez:
Added some units tests. LGTM now :+1:

PR #151, opened 2016-06-28
Feature/mortar 213

hkekeli:
https://jira.nationalgeographic.com/browse/MORTAR-213

git checkout
gulp serve 
http://localhost:3000/icons


xiwcx:
the code samples should be in separate columns with their visual examples

PR #156, opened 2016-06-30
[feature/AEM-5364] Create CTA Button/Create Icon Component

bpearsoncse:
This encompasses both stories as the AC is tightly coupled:
https://jira.nationalgeographic.com/browse/AEM-5406
https://jira.nationalgeographic.com/browse/AEM-5364

TO TEST:
Pull feature/AEM-5364
Run `gulp serve`
Change values in app/modules/button/docs.html && app/modules/icon/docs.html
Verify changes according to AC


app/modules/button/scripts/CTAButtonPestle.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+import React from 'react';
+import * as ReactDOM from 'react-dom';
+import {Pestle, Module} from '../../../scripts/pestle/main.js';
```
krlospelaez: @bpearsoncse Would you mind update your branch with master? There's a new way to import Pestle. Now you can use:
```javascript
import {Pestle, Module} from '@natgeo/mortar-pestle'
```

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
...
+    position: absolute;
+    top: 44%;
+    left: 50%;
+    width: auto;
+    height: auto;
+    -webkit-transform: translate(-50%,-50%);
+    -moz-transform: translate(-50%,-50%);
+    -ms-transform: translate(-50%,-50%);
+    -o-transform: translate(-50%,-50%);
+    transform: translate(-50%,-50%);
```
xiwcx: compilation is happening through [https://github.com/postcss/autoprefixer](autoprefixer), you don't need to worry about vendor prefixes

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
...
+  a.mt_btn {
+    color: inherit;
+    text-decoration: none;
+  }
+  .mt_btn{
+    height: 55px;
+  }
+  //Center icon and text within button, or center link if no icon
+  .mt_iconandlabel--horizontal, .mt_iconandlabel--vertical {
+    position: absolute;
```
xiwcx: attributes should be alphabetical, this will actually break the build (strict linting!). you can run `npm test` to make sure your code passes before PRs.

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
...
+  //Override for user-agent defaults link button text/icon
+  a.mt_btn {
+    color: inherit;
+    text-decoration: none;
+  }
+  .mt_btn{
+    height: 55px;
+  }
+  //Center icon and text within button, or center link if no icon
+  .mt_iconandlabel--horizontal, .mt_iconandlabel--vertical {
```
xiwcx: please one selector per line

```scss
.mt_iconandlabel--horizontal, 
.mt_iconandlabel--vertical {
```

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
+div[data-pestle-module="CTAButton"]{
```
xiwcx: little lost here, ideally the buttons shouldn't need special styles just for the CTA version. if they do and there's no way around it, then this should be a variation:

```scss
.mt_btn {...}

.mt_btn--cta { ... }
```

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
...
+  //Override for user-agent defaults link button text/icon
+  a.mt_btn {
```
xiwcx: a bird's eye view is best, if it doesn't make sense to just put these styles on the normal `.mt_btn` class then it should look like the following:

```scss
.mt_btn {
   ...

   a& {
     ...
  }
}
```

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,23 @@
...
+    color: inherit;
+    text-decoration: none;
+  }
+  .mt_btn{
+    height: 55px;
```
xiwcx: maybe a min-height here in the unfortunate event that the text bumps to a second row?

app/modules/button/_styles.scss
bpearsoncse: this was to just specifically override the default <a> color when the type is "link". Although, I may have gotten a bit overzealous in my specificity...

app/modules/button/scripts/CTAButtonPestle.js
bpearsoncse: yep, will update now

app/modules/button/_styles.scss
bpearsoncse: great! will delete

app/modules/button/_styles.scss
bpearsoncse: GTK, passing now

app/modules/button/scripts/CTAButton.jsx
```diff
@@ -0,0 +1,157 @@
...
+import Icon from "../../icon/scripts/icon.jsx";
+
+class CTAButton extends Component {
+
+  render() {
+    this.isTextLink = this.props.type === 'link';
+    this.isSubmit = this.props.type === 'submit';
+    this.isReset = this.props.type === 'reset';
+
+    let attrs = {
```
xiwcx: there are a few `let`s that seem like they could be `const`s, any reason behind the decision?

app/modules/button/scripts/CTAButton.jsx
```diff
@@ -0,0 +1,157 @@
...
+          let lastEl = index < array.length - 1;
+          terms += lastEl ? element + "+" : element;
+        }
+        termsArr.forEach(concatTerms);
+        attrs.href = attrs.href + "?" + "utm_source=" + this.props.link.trackingCodes.utmSource + "&utm_medium=" + this.props.link.trackingCodes.utmMedium +
+          "&utm_term=" + terms + "&utm_content=" + this.props.link.trackingCodes.utmContent + "&utm_campaign=" + this.props.link.trackingCodes.utmCampaign;
+      }
+    }
+
+    if (this.props.inactive) {
```
xiwcx: i've already got the [classnames](https://www.npmjs.com/package/classnames) package referenced elsewhere, may as well do it here to clean this up a 'lil.

app/modules/button/_styles.scss
bpearsoncse: 👍

app/modules/button/scripts/CTAButton.jsx
bpearsoncse: I'm sure there was a reason, although obviously not a good one 😜. looks like I can change all but a couple to `const`

app/modules/button/scripts/CTAButton.jsx
bpearsoncse: I played around with this quite a bit, but given the number of conditions and the dynamic nature of the style variable, I found it ended up quite complex and hard to read. Granted, it is getting late and maybe this could be easily simplified, but this is what I came up with:
`const buttonStyle = this.props.style;
const btnClasses = classNames({
  'mt_btn mt_fullwidth' : true,
  'mt_btn-naked' : this.isTextLink || this.props.style === "naked",
  'mt_btn-naked--inactive' : this.isTextLink && this.props.inactive,
  'mt_btn-naked--reversed' : this.isTextLink && this.props.inverse && !this.props.inactive,
  [`mt_btn-${buttonStyle}`] : !this.isTextLink && !this.props.inactive && !this.props.inverse,
  [`mt_btn-${buttonStyle}--inactive`] : !this.isTextLink && this.props.inactive,
  [`mt_btn-${buttonStyle}--reversed`] : this.props.inverse &&
                                        (this.props.style === "secondary" || this.props.style === "naked") &&
                                        !this.props.inactive && !this.isTextLink
});

const attrs = {
  className: btnClasses,
  onClick: (this.props.authorMode || this.props.inactive) ? null : this.props.onClick,
  onFocus: (this.props.authorMode || this.props.inactive) ? null : this.props.onFocus,
  onBlur: (this.props.authorMode || this.props.inactive) ? null : this.props.onBlur
};`
Going to leave my original block in there for now. I am open to suggestions if you think the classnames approach can be simplified.

bpearsoncse:
All changes made per comments with the exception of last. Please see my reply. Passing back to you @xiwcx

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,25 @@
...
+a.mt_btn {
+  color: inherit;
+  text-decoration: none;
+  &-naked--reversed,
+  &-naked--inactive {
+    color: white;
+  }
+}
+
+.mt_btn {
```
xiwcx: sorry, not trying to be a pain, nearly did this for you, but i should probably just keep trying until my nonsense is coherent. altogether this should be:

```scss
.mt_btn {
  ...

  a.& { // this gives a bird's eye view of what happens to .mt_btn in all of it's permutations
  ...

  &--naked--reversed, //double dashes for both naked and reversed, since they're both modifications and neither is a specific element within a block
  &--naked--inactive {
    ...
  }
}
```

Also, I ask this genuinely as I haven't tested myself, but can any harm come of omitting the `a.&` part altogether? Does `color: inherit; text-decoration: none;` have any negative effect on non link buttons?

bpearsoncse: I get: `Error: property "a" must be followed by a ':'` when I try to use the symbol in that manner. 
also, it does have a negative effect on both non-link and link buttons without the specificity.

bpearsoncse: with regard to the double dashes, I was following the styles in http://style-guide.nationalgeographic.com/latest/site/patterns/ which currently use single dashes.

xiwcx: my bad, there's a typo it's just `a&` that extra dot is superfluous. my bad about the button classes, those need to get corrected as well.

xiwcx: got it, but it's possible this is needlessly verbose. feel free to leave as is.

```scss
.mt_btn {
  color: red;

  @at-root{
    a#{&} { // this gives a bird's eye view of what happens to .mt_btn in all of it's permutations
    color: blue;
  
      &--naked--reversed, //double dashes for both naked and reversed, since they're both modifications and neither is a specific element within a block
      &--naked--inactive {
        color: green;
      }
    }
  }
}
```

app/modules/button/scripts/CTAButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+class CTAButton extends Component {
+
+  render() {
+    this.isTextLink = this.props.type === 'link';
+    this.isSubmit = this.props.type === 'submit';
+    this.isReset = this.props.type === 'reset';
+
+    const attrs = {
+      className: 'mt_btn mt_fullwidth ' + 'mt_btn-' + ((this.isTextLink || this.props.style === "naked") ? 'naked' : this.props.style),
+      onClick: (this.props.authorMode || this.props.inactive) ? null : this.props.onClick,
```
krlospelaez: @bpearsoncse I don't think we should tie AEM authorMode on the component. It'll be useless on other platforms. I think just leaving inactive props is enough. You can add authorMode logic on AEM sites specifically IMO.

app/modules/button/scripts/CTAButton.jsx
```diff
@@ -0,0 +1,149 @@
+'use strict';
+
+import React, { Component, PropTypes }  from 'react';
+import Icon from "../../icon/scripts/icon.jsx";
```
krlospelaez: Please change ``ìcon.jsx``` to ```Icon.jsx```. It isn't building on linux :(

bpearsoncse:
Back to you guys @xiwcx @krlospelaez . To note, I did not change the "naked" style as it seems consistent with the other element-level styles such as "default" and "secondary", whereas only "success" and "error" classes seem to be modifiers (double-dash) to me. I would be happy to change if you disagree, but that was my reasoning behind not making the change.

xiwcx:
hey @bpearsoncse, you may have missed it, but i actually already mentioned that the styles you're referencing are a mistake that slipped through the cracks. those will actually be updated to reflect the correct syntax as well.

app/modules/button/_styles.scss
```diff
@@ -0,0 +1,22 @@
...
+  color: inherit;
+  text-decoration: none;
+  &-naked--reversed{
+    color: white;
+  }
+  &-naked--inactive span,
+  &-naked--inactive svg,
+  &-naked--inactive:hover span,
+  &-naked--inactive:hover svg{
+    color: rgba(179,179,179,.4);
```
xiwcx: wait, is this new? is this a new color?

bpearsoncse: not a new color, just the inactive text color to override user agent default when nothing to inherit from

bpearsoncse: ^^ sorry about that, I must have missed that comment. will change that style now.

bpearsoncse:
@xiwcx changed class name and updated docs

krlospelaez:
@bpearsoncse Here's the approach (example) I mentioned you. https://gist.github.com/krlospelaez/46af8c5374eac0bbe1b327a5092fe751

bpearsoncse:
@krlospelaez - I used most of what you suggested, with the exception of the classnames module, as the display logic was already handled with my solution and I didn't want to have to recreate it using the module (namely the conditions to override style if type is link). This should be ready for final review now.

bpearsoncse:
@xiwcx passing back for final review

krlospelaez:
LGTM 👍  Great job :)

PR #158, opened 2016-07-01
[MORTAR-252] Improved the build to expose modules, svg and css

krlospelaez:
Test each of this commands:

- `gulp test`
- `gulp build`
- `gulp build -p`
- `gulp serve`


gulp/tasks/styles.js
```diff
@@ -7,6 +7,11 @@ const autoprefixer = require('gulp-autoprefixer');
 const sourcemaps = require('gulp-sourcemaps');
 const cleanCSS = require('gulp-clean-css');
 
+const docStylesSrc = 'app/site/_styles/*.scss';
```
xiwcx: would you mind abstracting this in to another file? gulp/paths.js

```js
module exports {
  docStylesSrc: 'app/styles/*.scss',
  ...
}
```

i've been meaning to do that for a while, that way we can use/change them across all the files.

xiwcx:
one note then this is good to go

gulp/tasks/styles.js
krlospelaez: Great idea, I'll do it

PR #160, opened 2016-07-05
make composable <SocialButton />, add google +

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-223

* check that the google plus button is present and works


app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -67,22 +67,38 @@ class MTFacebook extends Component {
...
+  'google-plus'
+]
+
+MTSharing.defaultProps = {
+  socialSites: socialSites
+}
+
 MTSharing.propTypes = {
   url: PropTypes.string,
+  socialSites: PropTypes.arrayOf(PropTypes.oneOf(socialSites)),
```
xiwcx: @krlospelaez, i did this, but it seems to be ignoring this if i give it unapproved values. any ideas?

app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -3,62 +3,62 @@
...
-    "status = 1, width = " + windowWidth
-      + ", height = " + windowHeight
-      + ", left = " + windowLeft
-      + ", top = " + windowTop
-      + ", resizable = 0"
-    );
-}
-
-class MTTwitter extends Component {
+class MTSocialButton extends Component {
```
xiwcx: @chuckcarpenter, there, composable, happy?

app/modules/sharing/scripts/MTSharing.jsx
krlospelaez: That kind of validation is not possible without creating a custom validator. There's another wait but it requires changing the structure from array to object. It would be something like:

```javascript
socialSites: PropTypes.shape({
  "facebook": PropTypes.bool,
  "twitter": PropTypes.bool,
  "google-plus": PropTypes.bool
})
```

krlospelaez:
@xiwcx Did you see AC #3. I guess you need to include this script https://apis.google.com/js/platform.js

xiwcx:
@krlospelaez we abandoned the scripts. it's much more efficient to pass the href rather than load a library (at least five) for each social site. @asabaylus ?

asabaylus:
im expecting that Welch knows this better than I do. I'm fine with eliminating that AC

krlospelaez:
So it looks fine to me 👍

PR #163, opened 2016-07-08
[MORTAR-257] Updated Pestle to handles escaped data

krlospelaez:
**Improvement**

This PR introduce a new way to pass data inside modules. Before you pass data as follows:

```html
<div
  data-pestle-module="MyModule"
  data-pestle-options='{
     "param1": "value1",
     "param2": "value2"
  }'>
</div>
```
The above way had some issues when single or double quotes were included even if json was well formed. To fix this issue pestle now receives the json object inside a script tag as showed below:

```html
<div data-pestle-module="MyModule">
  <script type="text/json" data-pestle-options>
  {
    "param1": "value1",
    "param2": "value2"
  }
  </script>
</div>
```

**Issue Evidence - Single Quotes **

Here you can see how a single quote ( ' ) affects the markup, closing the html data-peslte-options attribute too early which causes a malformed html & json object.

```html
<div
  data-pestle-module="MyModule"
  data-pestle-options='{
     "param1": "Carlos' PR",
     "param2": "value2"
  }'>
</div>
```

Using new approach single quotes are not creating a malformed json.

```html
<div data-pestle-module="MyModule">
  <script type="text/json" data-pestle-options>
  {
    "param1": "Carlos' PR",
    "param2": "value2"
  }
  </script>
</div>
```

**Issue Evidence - Double Quotes**

There're some cases where you want to include double quotes ( " ), if the json object is created by hand, it's highly probably you get a malformed json object and/or html. This isn't a Pestle issue but a json format one. So it affects the old and new way for passing data to modules.

Old way
```html
<div
  data-pestle-module="MyModule"
  data-pestle-options='{
     "param1": "They said "hello people"",
     "param2": "value2"
  }'>
</div>
```

New way
```html
<div data-pestle-module="MyModule">
  <script type="text/json" data-pestle-options>
  {
    "param1": "They said "hello people"",
    "param2": "value2"
  }
  </script>
</div>
```

So to fix this problem it's a better practice using a json encoder to ensure a well formed json. However if there's not option and you have to create a json object by hand, you can follow the next guide recommendations http://www.thorntech.com/2012/07/4-things-you-must-do-when-putting-html-in-json/

To test:

- Run `gulp serve`
- Go to components menu
- Verify all components are working correctly.


xiwcx:
OK, so I know that @benjachapman and @bpearsoncse are refactoring the slider react component so that it just swallows whatever markup is on the page. What happens when we see:

```html
<div data-pestle-module="Slider">
  <script type="text/json" data-pestle-options>
  {
    "param1": "value1",
    "param2": "value2"
  }
  </script>
  <div><img src="asfd"></div>
  <div><img src="asfd"></div>
  <div><img src="asfd"></div>
  <div><img src="asfd"></div>
</div>
```

krlospelaez:
@xiwcx The approach is similar as before. If you want to add that html you should pass it as a property inside the json object.

```html
<div data-pestle-module="Slider">
  <script type="text/json" data-pestle-options>
  {
    "param1": "value1",
    "param2": "value2",
    "param3": "<div><img src=\"asfd\"></div><div><img src=\"asfd\"></div><div><img src=\"asfd\"></div><div><img src=\"asfd\"></div>"
  }
  </script>
</div>
```

bpearsoncse:
@krlospelaez I am getting errors when passing the divs as strings to the slider, specifically `TypeError: Cannot read property 'length' of undefined` 
this is what I am doing in the refactored gallery
```
'use strict';

import React, { Component, PropTypes } from 'react';
import SliderComponent from '../../slider/scripts/MTSlider.jsx';

class MTPhotoGallery extends Component {

  render() {
    return (
      <SliderComponent>
        {PropTypes.markup}
      </SliderComponent>
    );
  }
}

MTPhotoGallery.propTypes = {
  markup: PropTypes.string
}

export default MTPhotoGallery
```
it seems like some sort of conversion will need to take place, as it seems that react-slick is not parsing the string correctly as child elements.

PR #166, opened 2016-07-11
Feature/mortar 231

xiwcx:
https://jira.nationalgeographic.com/secure/RapidBoard.jspa?rapidView=308&view=detail&selectedIssue=MORTAR-231

* `gulp serve`
* make sure the email button exists and opens your email client with the correct text
* make sure tests pass

### DIFF
--- a/app/modules/sharing/scripts/MTSharing.jsx
+++ b/app/modules/sharing/scripts/MTSharing.jsx
@@ -3,84 +3,12 @@
 import React, { Component, PropTypes }  from 'react';
 import {Pestle} from '@natgeo/mortar-pestle';
 import urlEncode from 'urlencode';
-
-class MTSocialButton extends Component {
-  constructor(props) {
-    super(props);
-    this.handleClick = this.handleClick.bind(this);
-  }
-
-  openShareWindow(site, url, title) {
-    const encodedUrl = urlEncode(url);
-    const siteAttr = {
-      facebook: {
-        h: 279,
-        url: "http://www.facebook.com/share.php?u=" + encodedUrl,
-        w: 575,
-        windowTitle: 'Facebook'
-      },
-      'google-plus': {
-        h: 647,
-        url: "https://plus.google.com/share?url=" + encodedUrl,
-        w: 500,
-        windowTitle: 'Google Plus'
-      },
-      linkedin: {
-        h: 425,
-        url: "https://www.linkedin.com/shareArticle?mini=true&url=nationalgeographic.com" + encodedUrl,
-        w: 450,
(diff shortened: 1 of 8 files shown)


app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -3,118 +3,70 @@
 import React, { Component, PropTypes }  from 'react';
 import {Pestle} from '@natgeo/mortar-pestle';
 import urlEncode from 'urlencode';
+import MTSocialButton from './MTSocialButton.jsx'
```
krlospelaez: I like this approach :+1:

app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -3,118 +3,70 @@
...
-class MTSocialButton extends Component {
-  constructor(props) {
-    super(props);
-    this.handleClick = this.handleClick.bind(this);
-  }
+class MTSharing extends Component {
+  render() {
+    // this looks wacky, but the strings need to be URL encoded
+    // while the API remains unencoded
+    const emailFormat = 'mailto:?subject='
```
krlospelaez: Why not keep this inside MTSocialButton? Try to keep all buttons logic in one place.

krlospelaez:
The branch has some merge conflicts

krlospelaez:
I tested it on Mozilla Thunderbird Mail but seems like it isn't accept HTML on the body.

image

app/modules/sharing/scripts/MTSharing.jsx
xiwcx: because email works completely differently than the rest of the buttons. doesn't seem to make sense to force the logic if there's no click event. while not the cleanest, this does feel like the most efficient way to do it. i could break it in to it's own component, but it's really just returning markup and nothing else.

krlospelaez: You're right! It shouldn't be a social button. Would you mind create another component/file with the email logic on it (MTEmailButton maybe?). It'll keep a single responsibility on each component.

So MTSharing is just responsible to add and remove the email and social buttons and MTEmailButton is responsible to create an email button :)

xiwcx:
@krlospelaez addressed merge conflicts and comments. not sure what to do about thunderbird, i'm following the `mailto:` API and the html works in most places. this is probably a decision for natalie or kevin.

krlospelaez:
@xiwcx https://github.com/natgeo/mortar/pull/166/files#r70453297

krlospelaez:
LGTM :+1:

PR #167, opened 2016-07-11
Remove focus from inactive buttons 

hkekeli:
https://jira.nationalgeographic.com/browse/MORTAR-241

git checkout
gul serve
Focus is removed from inactive buttons so you can no longer tab through. http://localhost:3000/patterns


xiwcx:
@benjachapman @bpearsoncse would it make sense to add a disabled option to the CTA module?

bpearsoncse:
@xiwcx currently there is an inactive option which should achieve the same effect

PR #168, opened 2016-07-11
SYTO-647: Set up Slider component to support a content-agnostic appro…

bpearsoncse:
…ach, add dummy Photo Gallery component to demonstrate

https://jira.nationalgeographic.com/browse/SYTO-647


xiwcx:
before I even go full review on this, let's make sure this approach is synced up with @krlospelaez's work on #163

app/modules/photoGallery/docs.html
```diff
@@ -0,0 +1,14 @@
+<div class="pagewrapper mt_row mt_col-gut-sm mt_col-md-gut-md mt_col-lg-gut-lg">
+  <div class="mt_col-lg-3 mt_col-sm-12 mt_row mt_col-gut-sm mt_col-md-gut-md mt_col-lg-gut-lg mt_subh4">
```
krlospelaez: @bpearsoncse Remember mortar classes prefix has changed. So you need to change `mt` prefix to `mt2` e.g. `mt2_col-lg-3`

app/modules/slider/scripts/MTSlider.jsx
```diff
@@ -72,30 +72,43 @@ class MTSlider extends Component {
...
+    let slides;
 
-      return <div key={i}>{slideMarkup}</div>;
-    });
+    /*
+     * if a "slides" prop is provided, a set of slides of a supported type will be rendered.
+     * Otherwise, the component will render any children passed to the component,
+     * arranging sibling elements as slides
+     */
+    if(props.slides) {
```
krlospelaez: IMO slides property must be removed, we should keep Slider API consistent and just offer a single way to add slides (children prop). What do you think? @xiwcx @asabaylus @benjachapman

bpearsoncse:
@xiwcx @asabaylus - after further discussion with @krlospelaez it is our opinion that a content agnostic approach would require some less-than-desirable architectural decisions, namely the method of injecting markup through a property in Pestle. We propose that we continue with the configuration paradigm, and expand the functionality as additional formats are required. An example can be seen here: https://gist.github.com/krlospelaez/b34ffa4d9cfd96bc740fc1d2c0471a67 This approach will work well as I port over the existing Photo Gallery as well, as it uses a very similar approach to handling the slides.

asabaylus:
It sounds like you've all done you're due diligence.

PR #174, opened 2016-07-19
remove html from email body

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-231

* click the email button in the sharing component
* make sure there are no unescaped HTML tags in the body of the email


hkekeli:
Seeing this when I send it to myself. Is this expected behavior? This is happening in outlook /web on a mac. 
outlook web mac

xiwcx:
@hkekeli can you try a different browser? it seem like maybe you have a plugin installed that is messing with the URLs?

hkekeli:
@xiwcx Nope. Otherwise the initial state looks fine. Passing since the ticket was only concerned with the initial state.

PR #177, opened 2016-07-20
get rid of React console errors, false positives in unit tests

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-258

* pull branch
* run `gulp karma`
* see that the output is much cleaner


krlospelaez:
LGTM :+1:

PR #181, opened 2016-07-25
Feature/mortar 75

hkekeli:
https://jira.nationalgeographic.com/browse/MORTAR-75

Simplified code as much as I could. 

Instructions:

git checkout branch
gulp serve 
Visit http://localhost:3000/patterns
Make sure tooltip stays open on hover


app/modules/tooltips/_styles.scss
```diff
@@ -0,0 +1,35 @@
+.mt2_tooltip-container {
+  max-width: $mt2_unit * 18;
+
+  // Padding Needed to prevent long text rendering issues & keep toolip open on hover
+  padding-top: .0625rem;
```
xiwcx: that hard return threw me and i thought it was a separate block, let's just do this:

```scss
  max-width: $mt2_unit * 18;
  padding-top: .0625rem; // Padding Needed to prevent long text rendering issues & keep toolip open on hover
```

aside from this one thing, this is rock solid!

hkekeli: Updated.

PR #183, opened 2016-07-26
migrate modal from jquery to react

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-262

* see that the modal works as expected
* see that the modal correctly grabs markup from the page

questions:

- should there be an option to pass a string instead of grabbing markup?


hkekeli:
Modal works. 

Since the modal grabs the markup from the page, how well does this interact with other pestle components? For example, if the markup is a pestle component will it render via pestle before its HTML is grabbed? And if it the rendering of the child component is done via React, will there be issues b/c the component is essentially being moved in the DOM tree?

It would be nice if the modal component excluded the ‘launch modal’ button.  That way you could easily use it with other components. 

For example: 

```
<ModalComponent>
   <AnotherPestleComponent/>
 </ModalComponent>
```

Content can be passed in as child elements of the react component. This means you can even use React components like the Slider as children.

xiwcx:
@hkekeli all great questions and suggestions, but for now none of this (even the grabbing content) is in the AC, so we can create new tickets for this functionality and merge what we have

krlospelaez:
It's working fine now on my local.

image

PR #186, opened 2016-07-28
Create tooltip component

hkekeli:
https://jira.nationalgeographic.com/browse/SYTO-664

Instructions:
 
Tooltip component is created with the following directions: 
Top, right, bottom, left

git checkout branch
gulp serve 
http://localhost:3000/components


app/modules/tooltips/scripts/MTTooltip.jsx
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+import React, { Component, PropTypes }  from 'react';
+import {Pestle} from '@natgeo/mortar-pestle';
+import classNames from 'classnames';
+
+class Tooltip extends Component {
+  render() {
+    const tooltipClasses = classNames({
```
xiwcx: for `tooltipClasses` and `tooltipBtnClasses` there's no need to use the `classNames` utility since every key in your object is true every time. you can just set these directly on the `className=` property of the node. other than that, this looks great!!

PR #191, opened 2016-08-01
add truncation to social sharing component

xiwcx:
https://jira.nationalgeographic.com/browse/MORTAR-251

* browse to component page
* view the behavior of the vertical component
* for extra credit you can edit the values in the html of the example


app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -32,11 +60,30 @@ class MTSharing extends Component {
...
         />
       )
-    })
+    });
+    const closedContainerStyle = {
+      height: this.props.count > 0 && this.props.count < this.props.buttons.length && this.props.display === 'vertical'? ((this.props.count * 48) + 49) + 'px' : 'auto',
+      width: this.props.count > 0 && this.props.count < this.props.buttons.length && this.props.display === 'horizontal'? ((this.props.count * 48) + 49) + 'px' : 'auto'
+    };
+    const openContainerStyle = {
+      height: this.props.count > 0 && this.props.count < this.props.buttons.length && this.props.display === 'vertical'? ((this.props.buttons.length * 48) + 49) + 'px' : 'auto',
```
krlospelaez: Could be possible calculate width and height instead of using a fixed value? https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect could be useful on this case.

app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -7,12 +7,40 @@ import MTEmailButton from './MTEmailButton.jsx'
...
+  }
+
+  update(e) {
+    this.setState({
+      open: !this.state.open
+    })
+  }
+
+  escClose(e) {
+    if((e.charCode === 27 || e.keyCode === 27) && this.state.open === true) {
```
krlospelaez: Please, create a constant ESC_KEY = 27. It's easier to read what key you're referring to.

app/modules/sharing/scripts/MTSharing.jsx
alexrqs: this `((this.props.buttons.length * 48) + 49) + 'px'` looks like an great case for ES6 template strings.
```es6
// Interpolate variable bindings
var name = "Js Function", place = "bar";
`${name} walks into a ${place}?`
```

app/modules/sharing/_styles.scss
```diff
@@ -10,9 +25,41 @@
   background-color: transparent;
   border: 0;
   cursor: pointer;
-  padding: .75rem;
+  height: 48px;
+  min-height: 48px;
+  min-width: 48px;
+  padding: 12px;
+  width: 48px;
```
alexrqs: Note that max-width and width (same for height) shouldn’t be applied to the same element using the same unit, as one will override the other.

app/modules/sharing/scripts/MTSharing.jsx
```diff
@@ -32,11 +60,30 @@ class MTSharing extends Component {
           title={this.props.title}
         />
       )
-    })
+    });
+    const closedContainerStyle = {
+      height: this.props.count > 0 && this.props.count < this.props.buttons.length && this.props.display === 'vertical'? ((this.props.count * 48) + 49) + 'px' : 'auto',
```
pmourelle: Man, this looks sooo nasty. Setting it through CSS (without using magic numbers for pixels here) and maybe adding a class for layout (horiz/vert) wouldn't be better?

chuckcarpenter: You should just be calculating this in another var and making it DRY. ln 69 looks very similar.

PR #193, opened 2016-08-08
Apply max-width to tooltip

hkekeli:
https://jira.nationalgeographic.com/browse/SYTO-698

-Instructions

- git checkout
- gulp serve
- Visit  http://localhost:3000/components

Look for the following: 

- Tooltip content now supports a max-width
- Hover area is limited to the button. The tooltip is activated when you hover over the button.

### DIFF
--- a/app/modules/tooltips/_styles.scss
+++ b/app/modules/tooltips/_styles.scss
@@ -1,17 +1,17 @@
 $mt2_arrowheight: .625rem;
 
 .mt2_tooltip-container {
-  max-width: $mt2_unit * 3;
-  padding: .5rem;
   position: relative;
   transition: 400ms;
+  pointer-events: none;
+  margin: 10px;
 }
 
 .mt2_tooltip-btn {
   cursor: pointer;
-  display: block;
-  margin: 10px auto;
   padding: 1px 6px 0;
+  pointer-events: auto;
+  display: block;
 
   &:hover {
     color: $mt2_color-neutral-l;
@@ -23,11 +23,14 @@ $mt2_arrowheight: .625rem;
   position: absolute;
   text-align: center;
   z-index: $mt2_z-top;
+  pointer-events: auto;
+  max-width: 300px;
 
   &--top {
     bottom: 100%;
-    left: 50%;
     transform: translateX(-50%);
+    margin-left: 10px;
+    margin-bottom: 1rem;
 
     &::after {
       border: $mt2_arrowheight solid transparent;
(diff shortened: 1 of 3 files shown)


app/modules/tooltips/scripts/MTTooltip.jsx
```diff
@@ -38,6 +38,14 @@ class Tooltip extends Component {
     }
   }
 
+  componentDidUpdate(){
+    if(this.props.placement === "left" && this.state.visible){
+      let width =  window.getComputedStyle(this.refs.tooltipContainer).width;
+      width = -parseFloat(width) - 15;
```
alexrqs: `15` sounds very random, any reason? is some kind of constant or default value?

app/modules/tooltips/_styles.scss
```diff
@@ -59,9 +62,10 @@ $mt2_arrowheight: .625rem;
   }
 
   &--bottom {
-    right: 50%;
     top: 100%;
-    transform: translateX(50%);
+    transform: translateX(-50%);
+    margin-top: 1rem;
+    margin-left: 10px;
```
alexrqs: any reason why you dont store this `10px` on a variable, looks like you are using this value more than once. (same for margin top), also any reason to mix rem and px on the same prop?

alexrqs: remember sort A-Z for css

app/modules/tooltips/scripts/MTTooltip.jsx
pmourelle: Is that referred to the arrow's width? Putting it in a variable would help a lot readability and future changes ;)

app/modules/tooltips/_styles.scss
pmourelle: I second that.

app/modules/tooltips/scripts/MTTooltip.jsx
hkekeli: I just eyeballed the measure there to move the arrow further to the left.

pmourelle:
Testing it locally, only blue and green sections react to hover and triggers tooltip showing. 
Is this ok? (testing instructions aren't too specific on what to test)
image

hkekeli:
@pmourelle did you pull the latest changes on this branch? the tooltip should ***only*** activate when you're hovering over the button. The margin and the padding of that container shouldn't be included.

app/modules/tooltips/scripts/MTTooltip.jsx
hkekeli: this is what happens otherwise screen shot 2016-08-09 at 2 38 28 pm

pmourelle:
Nice changes.
ToolTip also works as it says.
PR OK!

alexrqs:
good job, PR OK!!

PR #194, opened 2016-08-09
SYTO-730: FE: Elevate AEM's video component to Mortar

pgurley:
https://jira.nationalgeographic.com/browse/SYTO-730

This story elevates AEM's "Embedded Video" component to Mortar

To test:
Pull branch SYTO-730
From within mortar directory, run `npm install`
From within mortar directory, run `npm test` to verify component's tests pass
From within mortar directory, run `gulp serve`
Verify component's functionality at http://192.168.1.128.xip.io:3000/components/

### DIFF
--- a/app/modules/video/_styles.scss
+++ b/app/modules/video/_styles.scss
@@ -1,4 +1,74 @@
-.embedded-video  {
-  /* temporary troubleshooting */
-  min-height: 375px;
+$desktop       : "(min-width: 70em)";
+
+.embedded-video {
+    position: relative;
+    padding-bottom: 56.25%; // 16:9
+    height: 0;
+    &:before {
+        content: " ";
+        display: block;
+        position: absolute;
+        top: 1px;
+        left: 1px;
+        right: 1px;
+        bottom: 1px;
+        background: #f2f2f2 url('/etc/designs/platform/v2/images/logos/ng-logo-small.svg');
+        background-repeat: no-repeat;
+        background-size: 60px;
+        background-position: center center;
+    }
+    &__player-wrapper {
+        position: absolute;
+        top: 0;
+        height: 100%;
+        width: 100%;
+    }
+    &__init {
+        @media #{$desktop} {
+            margin-bottom: 55px;
+        }
+    }
 }
+
+.embedded-video__player {
+
+    .embedded-video & {
+        position: relative;
+        width: 100%;
+        height: 100%;
+    }
+
+
+    .align-right {
+        float: right;
+        margin: 0 1rem;
+    }
+
+    .align-left {
+        float: left;
+        margin: 0 1rem;
+    }
+
+    a, td, th {
+        border: none;
+    }
+
+    .playerDiv,video,iframe {
+        width: 100%;
+    }
+
+    .inline {
+      position: relative;
+      padding-bottom: 56.25%; // 16:9
+      height: 0;
+        iframe {
+            position: absolute;
+            padding-bottom: 56.25%;
+            height: 100%;
+            width: 100%;
+            top: 0;
+            left: 0;
+        }
+    }
+}
+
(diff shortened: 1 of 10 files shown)


app/modules/video/scripts/video.jsx
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+import React, { Component, PropTypes }  from 'react';
```
alexrqs: `PropTypes` as static declaration for the class is not used, any reason to import ?

app/modules/video/scripts/VideoPestle.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
alexrqs: Please remove this http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code (based on section 10.2.1)

app/modules/video/scripts/video.jsx
```diff
@@ -0,0 +1,89 @@
...
+  handleOnLoad() {
+    // if we are not using editmode
+    if (this.props.isEditMode !== true) {
+      this.createPlayer();
+    }
+  }
+
+  render() {
+    // LazyLoad will defer PDK initialization until component is within 200px of the visible viewport
+    if (this.props.lazyLoad === true) {
```
alexrqs: `this.props` is used more than 5 times on this method, consider destructure the variables to improve readability.

app/modules/video/scripts/VideoPestle.js
pgurley: It seems like every module in the codebase makes that declaration, and it's called out in the style guide
https://github.com/natgeo/javascript#modules-1

Do we really want to leave it out on this one component?

app/modules/video/scripts/video.jsx
pgurley: I don't think there's a real need to import it. I was following the existing conventions being used in other modules. I'll update

app/modules/video/scripts/video.jsx
pgurley: @alexrqs I pushed an update with the suggested variable assignments destructued

app/modules/video/scripts/video.jsx
```diff
@@ -0,0 +1,97 @@
...
+  /**
+  * returns the options objects overriding any existing defaults
+  */
+  shapeModel() {
+    return _defaultsDeep(this.props.model, {
+      autoload: true,
+      type: 'fng',
+      skinName: (this.props.model.overlayPlayButton) ? 'glass-ngs' : 'glass-ngs-no-overlay',
+      layoutOptions: {
+        tpShare: (this.props.model.sharing !== undefined) ? this.props.model.sharing : true
```
sebasrgl: Please change all places where you set this property for ``tpShareFNG`` since the FNG player use that property to enable/disable the share component

pgurley: This is updated

app/modules/video/_styles.scss
```diff
@@ -0,0 +1,19 @@
+.mt2_video {
+  &__player {
```
xiwcx: i know it's weird, but we format BEM a little different: `blockname-elementname-modifier name`

which would make these `&-player` and `&-playerwrapper`

app/modules/video/scripts/VideoPestle.js
xiwcx: agreed with @pgurley, if we want to change it we need to change it on the whole which is out of the purview of this ticket.

app/modules/video/_styles.scss
pgurley: Good to know. @xiwcx would it be preferable to name the wrapper just `-wrapper`?

app/modules/video/scripts/VideoPestle.js
alexrqs: Good point guys, thanks for bringing that out :wink: fun thing that reference, @pgurley, is to _ES5.1 function expression_ and this file is an _ES6 Module Declaration_ feel free to read the specification of the lenguage you are actually using http://www.ecma-international.org/ecma-262/6.0/

PR #198, opened 2016-08-11
[SYTO-727][SYTO-728] Hello World content package

krlospelaez:
[SYTO-727] Create Hello World content package
[SYTO-728] Test Example Page in Mortar

I created a single PR for these 2 tickets because there's no way to test [SYTO-727] without [SYTO-728] and vice versa.

To test:

- `git checkout feature/SYTO-728`
- `gulp serve`
- Go to Content Packages


bpearsoncse:
This looks great! 👍

PR #203, opened 2016-08-15
[feature/MORTAR-275-4] fix duration display bug

bpearsoncse:
To test:
```
checkout feature/MORTAR-275-4
go to: http://10.0.0.153.xip.io:3000/components/promo
change type to video
make sure is a value below the kicker
```


asabaylus:
I dig it!

PR #205, opened 2016-08-17
[SYTO-729] Build video playlist component in Web Tools/Mortar (Elevate AEM's component)

sebasrgl:
https://jira.natgeo.com/browse/SYTO-729

#### How to test
1. ``npm i``
2. ``gulp serve``
3. Go to http://localhost:3000/components/
4. Check the video playlist's functionality 
5. :elephant:


app/modules/videoplaylist/scripts/VideoPlaylist.jsx
```diff
@@ -0,0 +1,258 @@
...
+      //This should set the proper url for the video player instance and start playing it
+      ngsPlayer.api.set(url, this.playerInstanceName);
+    } else {
+      //set flag to use in "OnLoadReleaseUrl" call back, so we know to call play on player instance
+      this.manualPlay = true;
+      ngsPlayer.api.unload(this.playerInstanceName);
+      ngsPlayer.api.load(url, this.playerInstanceName);
+    }
+
+    this.showHiddenThumb();
```
pgurley: I think this needs to be moved to the updateVideo method or the method otherwise updated. Since the state hasn't changed at the point this gets called, the current index is off by one. The result is that as you get to the last thumbnail, which should get scrolled into view you don't. Likewise, when that video finishes playing the first thumbnail should scroll into view

app/modules/videoplaylist/docs.html
```diff
@@ -0,0 +1,66 @@
...
+      }
+      </script>
+    </div>
+  </div>
+  <div class="mt2_col-lg-10 mt2_col-lg-offset-2 mt2_col-sm-12 mt2_mt2_row-gut-4">
+    <p class="mt2_subh4">This is the way to create a Video Playlist with autocontinue playback
+    support. Autocontinue property is by default enabled. </p>
+    <div class="codesnippet-wrapper">
+      <pre class="codesnippet"><code class="language-markup">
+        &#x3C;div data-pestle-module=&#x22;Video&#x22;&#x3E;
```
pgurley: This should be `data-pestle-module="VideoPlaylist"`

app/modules/videoplaylist/docs.html
```diff
@@ -0,0 +1,66 @@
...
+      }
+      </script>
+    </div>
+  </div>
+  <div class="mt2_col-lg-10 mt2_col-lg-offset-2 mt2_col-sm-12 mt2_mt2_row-gut-4">
+    <p class="mt2_subh4">This is the way to create a Video Playlist without autocontinue playback
+    support.</p>
+    <div class="codesnippet-wrapper">
+      <pre class="codesnippet"><code class="language-markup">
+        &#x3C;div data-pestle-module=&#x22;Video&#x22;&#x3E;
```
pgurley: This should be `data-pestle-module="VideoPlaylist`

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
```diff
@@ -0,0 +1,258 @@
+import React, {Component} from 'react';
+import Velocity from 'velocity-animate';
+import Truncate from 'react-truncate';
```
pgurley: Is Truncate being used anywhere in this component?

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
```diff
@@ -0,0 +1,258 @@
...
+
+import VideoThumbnail from './VideoThumbnail.jsx';
+import VideoCaption from './VideoCaption.jsx';
+
+class VideoPlaylist extends Component {
+
+  constructor(props) {
+    super(props);
+    this.autoPlay = false;
+    this.playerInstanceName = 'platformPlaylistVideoPlayer_' + this.props.divID;
```
pgurley: Should we update this to use template strings?
`platformPlaylistVideoPlayer_${this.props.divID}`

app/modules/videoplaylist/scripts/VideoThumbnail.jsx
```diff
@@ -0,0 +1,50 @@
...
+          <Truncate lines={3} ellipsis={(<span>...</span>)}>
+            <div ref="videoTitle"  dangerouslySetInnerHTML={{__html: this.props.item.title}} />
+          </Truncate>
+        </a>
+      </div>
+    )
+  }
+}
+
+VideoThumbnail.PropTypes = {
```
pgurley: I believe the 'P' should be lower case here

`VideoThumbnail.propTypes`

app/modules/videoplaylist/scripts/VideoThumbnail.jsx
```diff
@@ -0,0 +1,50 @@
...
+  wrapperClass: React.PropTypes.string,
+  item: React.PropTypes.shape({
+    directLink: React.PropTypes.string,
+    title: React.PropTypes.string,
+    guid: React.PropTypes.string
+  }).isRequired
+}
```
pgurley: All of the props being used in the component aren't being validated. Should it be more like this?
```
VideoThumbnail.propTypes = {
  onClick: React.PropTypes.func,
  wrapperClass: React.PropTypes.string,
  item: React.PropTypes.shape({
    directLink: React.PropTypes.string,
    title: React.PropTypes.string,
    guid: React.PropTypes.string,
    thumbnail: React.PropTypes.string,
    path: React.PropTypes.string,
  }).isRequired,
};
```

app/modules/videoplaylist/scripts/VideoCaption.jsx
```diff
@@ -0,0 +1,24 @@
...
+          <span itemProp='headline' dangerouslySetInnerHTML={{__html: this.props.title}} />
+        </h3>
+        <div ref="abstract" className="multi-layout-promos__promo-dek">
+          <span itemProp='description' className="mt2_subh4" dangerouslySetInnerHTML={{__html: this.props.abstract}} />
+        </div>
+      </div>
+    )
+  }
+}
+
```
pgurley: We should probably validate the propTypes here

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
```diff
@@ -0,0 +1,258 @@
...
+  getBreakpoint() {
+    const body = document.querySelector('body');
+    return window.getComputedStyle(body, ':after').getPropertyValue('content').replace(/"/g, '').replace(/'/g, '') || 'not available';
+  }
+
+  updateVideo() {
+    if (this.state.currentVideoIndex < (this.props.dataModel.length - 1)) {
+      this.changeVideo(this.state.currentVideoIndex + 1);
+    } else if (this.state.currentVideoIndex === (this.props.dataModel.length - 1)) {
+      this.changeVideo(0);
```
pgurley: Hitting this block in the original version of the component was intended to stop playback when returning to the first video automatically. That logic doesn't seem to have been retained (the playlist keeps looping). I believe this is tied to the order in which things are firing now that updates are being handled through state changes. I'd think it'd be pretty easy to adapt that logic (maybe with autoPlay as a state var) and clean this up a little

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
sebasrgl: You are right, I'll remove

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
sebasrgl: I did the change sending the ``nextState`` parameter in order to keep the consistency among the functions. The expected result is the same due to the current index in the object is the proper in this timeline.

app/modules/videoplaylist/scripts/VideoCaption.jsx
sebasrgl: I put the validation.

app/modules/videoplaylist/scripts/VideoPlaylist.jsx
sebasrgl: The ``START_VIDEO_INDEX`` constant was created to store the value of the initial index, so we can validate if the playlist did a loop and stop the playback.

pgurley:
Looks good to me. Merged!

PR #215, opened 2016-08-23
Fix merge conflict on tooltips

hkekeli:
No ticket. Max-width works again and this no longer occurs.


benjachapman:
Looks good to me - thanks!

PR #216, opened 2016-08-23
[feature/SYTO-756-2] Bug fix: Multiple URLs triggered by click events

bpearsoncse:
This is a bug fix QA found for https://jira.natgeo.com/browse/SYTO-756, in which when the user clicked on the kicker, the container link would trigger as well. This was due to the link being nested, despite having a higher z-index. This was refactored to not be nested, as well as updating unit tests and adding additional fields to the kicker on the test page.


benjachapman:
After settling a minor quibble or two, this LGTM 👍

PR #217, opened 2016-08-24
[SYTO-769] Hero With Two Rails component

benjachapman:
https://jira.natgeo.com/browse/SYTO-769

This work sets up the Hero With Two Rails content package component. It is intended to house other Story Card components, and have some of its own behavior/features.

The component is currently unstyled, and has no additional behavior other than spitting out its data model in the appropriate place.

Because this component will need to house parsyses in AEM, its structure is atypical from other Mortar components - it must be able to inject React content into pre-existing divs within the DOM, so as to preserve AEM's drag-and-drop authoring for child promos. To accomplish this, this work adds a "Portal" utility function, which allows one to target React components to specific places in the DOM. 

This unusual architecture also necessitated some light manipulation of how the Configurator pages have typically worked. 

Tests/configurator added as well.


pgurley:
This seems to behave as advertised. Tests all pass and nothing really jumps out to me. There are a couple of commented out lines that I'd like to see removed in future work.

PR #219, opened 2016-08-24
[SYTO-732] Create Heading classes for each font

hkekeli:
https://jira.natgeo.com/browse/SYTO-732

UI is included on the typography page for testing purposes.

Instruction:

- On the typography page, confirm that headings are responsive and the right fonts are being used.


app/styles/themes/default/vars/_typography.scss
```diff
@@ -120,6 +120,212 @@ $mt3_unit-quadruple: $mt3_unit * 4;
...
+}
+
+.mt3_heading--primary {
+  font-family: $mt3_haas-stack;
+  font-weight: 600;
+  letter-spacing: .05em;
+  line-height: 1.1;
+
+  h1 {
+    font-size: 1.875rem;
```
alexrqs: Hi @hkekeli I'm seen some repetition between the line https://github.com/natgeo/mortar/pull/219/files#diff-7f1fbde1e7f11b7e99bcf19301e28e8dR195 and this line, same apply for the rest of the tags.. I was wondering if you can abstract this on a mixin, funcion, variable or whichever way you decide, to centralize this only in one place instead hardcode the values here, that way when the value changes you only have to commit on one place, and the code will be a lot less verbose.

app/styles/themes/default/vars/_typography.scss
```diff
@@ -120,6 +120,212 @@ $mt3_unit-quadruple: $mt3_unit * 4;
 
 // Old Styles
 
+.mt3_heading--display {
```
pmourelle: Maybe the class has to be applied to `Hx` instead of its wrapper. That allows you to have just 3 selector definitions without all that nested hx whitin.
Summing to Alex's comment below, font sizes would be better to be handled using a mixin for ease legibility and future changes.

hkekeli: I thought these 3 classnames were substitutes for the `.mt3_article` class but with different fonts. See` article.scss`. I could be wrong because I struggle to how these would be used.

app/styles/themes/default/vars/_typography.scss
hkekeli: Yes, you’re right. Most of these were imported from the AC. I thought the values were all different but on a second look, there’s some repetition that could be avoided.

app/styles/themes/default/vars/_typography.scss
```diff
@@ -120,6 +120,212 @@ $mt3_unit-quadruple: $mt3_unit * 4;
...
+
+    @media #{$mt3_mq-lg} {
+      font-size: 1.475rem;
+    }
+  }
+
+  h6 {
+    font-size: 1.125rem;
+  }
+}
```
xiwcx: for the record, i hate this approach. we're tying presentation to markup and basically rendering the headings meaningless all while necessitating an unnecessary wrapper. these should all just be classes.

```html
<div class="mt3_heading--primary">
  <h2>text</h2>
<div>
```

this is anti-html


```html
<h3 class="mt3_heading--primary--h3">text</h3>
```

the above is better for flexibility, better for our html being at all meaningful and simply isn't that much more verbose

hkekeli: hey, I'm already working on something similar. See @pmourelle's comment on Jira. 

```css
h3 {
  &.mt3_heading-primary,
  &.mt3_heading-body {
    font-size: 1.575rem;

    @media #{$mt3_mq-md} {
      font-size: 1.875rem;
    }

    @media #{$mt3_mq-lg} {
      font-size: 2.175rem;
    }
  }

  &.mt3_heading-display {
      font-size: 2.1375rem;

      @media #{$mt3_mq-md} {
        font-size: $mt3_fs-med-lrg;
      }

      @media #{$mt3_mq-lg} {
        font-size: $mt3_fs-med-xlrg;
      }
    }
}
```

app/styles/themes/default/vars/_typography.scss
```diff
@@ -49,8 +46,212 @@ $mt3_unit-triple:    $mt3_unit * 3;
 $mt3_unit-quadruple: $mt3_unit * 4;
 
 
-// New Styles
 
+// New headings font-size variables
+// Naming scheme can be improved in future updates
+
+
+$mt3_h-sm1: 1.125rem;     // 18px
```
pmourelle: This is a no go on these vars.
Better to go as Alex suggested: having a base size for each font family, the sizes of headers can be calculated from it as they seem to have the same proportion. So, you'd have just 3 variables with sizes and you may calculate the rest of values from those.

app/styles/themes/default/vars/_typography.scss
```diff
@@ -49,8 +46,212 @@ $mt3_unit-triple:    $mt3_unit * 3;
...
+}
+
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  &.mt3_heading-primary {
```
pmourelle: I think we agreed before in jira comments that naming to use is: 'heading-1', 'heading-2' and 'heading-3' as current naming could create some confusion.

app/styles/themes/default/vars/_typography.scss
hkekeli: some of these sizes are consecutive

pmourelle:
Looks good! A couple of concerns that may be covered in a different ticket if needed:
- all that font sizes still not looking good to me. When making a change in future would be easier to change just one value and rest got calculated automatically.
- headers listed may need a code snippet next to them about how to use them or what differences are between each other.
Being that said (and to be taken care in new tickets if needed), this is ready to go. 🚢

app/styles/themes/default/vars/_typography.scss
```diff
@@ -49,8 +46,211 @@ $mt3_unit-triple:    $mt3_unit * 3;
...
+  }
+  &.mt3_heading-2 {
+    @extend %mt3_verlag-heading;
+  }
+  &.mt3_heading-3 {
+    @extend %mt3_georgia-heading;
+  }
+}
+
+h1 {
```
xiwcx: this is still too specific and goes against the way almost all classes are written in mortar. targeting html elements is needless and too specific. what happens when you want to use this style on a <span> or <button>?

pmourelle: @xiwcx you got cut in the html comment. How can get this be done? Adding the 'h1' as part of the class and leaving it as generic maybe?

app/styles/themes/default/vars/_typography.scss
```diff
@@ -46,29 +46,36 @@ $mt3_unit-triple:    $mt3_unit * 3;
...
 
 
-$mt3_h-sm1: 1.125rem;     // 18px
-$mt3_h-sm2: 1.275rem;     // 20px
-$mt3_h-sm3: 1.475rem;     // 23px
-$mt3_h-sm4: 1.575rem;     // 25px
-$mt3_h-sm5: 1.625rem;     // 26px
+$mt3_h-base: 1.125rem;
+
+$mt3_h-sm1: $mt3_h-base;
```
pmourelle: Better.
So this `$mt3_h-sm1` would be 19,875px?

hkekeli: 18px

app/styles/themes/default/vars/_typography.scss
xiwcx: perhaps `mt_heading--1--1` or  `mt_heading--1--sm`

hkekeli: how about `mt_heading--h1--1`? where the number at the end would reference the font being used?  (see latest update)  `mt_heading--1--1` could be confusing to use. Not sure of the thought process behind `mt_heading--1--sm`

hkekeli: @xiwcx Any thoughts on the above? ^

pmourelle:
LGTM!
Any further work needed (if any) would be out of current AC. Mergin'

PR #222, opened 2016-08-25
[SYTO-729] Build video playlist component in Web Tools/Mortar (Elevate AEM's component)

sebasrgl:
This PR is intended to fix some problems in the video playlist component
https://jira.natgeo.com/browse/SYTO-729

### How to test
1. npm i
2. gulp serve
3. Go to http://localhost:3000/components/
4. Check the videoplaylist component
5. Click on the first playlist's video with autocontinue true and be sure that video is playing.
6. Check all functionality with the playlist with autocontinue false.


app/modules/videoplaylist/scripts/VideoPlaylist.jsx
```diff
@@ -61,27 +62,29 @@ class VideoPlaylist extends Component {
   }
 
   handleEvent(e) {
+    if (!(ngsPlayer && ngsPlayer.api && typeof $pdk != 'undefined')) {
```
pgurley: Would it make sense to combine and the return below into one block, or does that make it too hard to read?

sebasrgl: I think is more readable like this, so, I prefer this way, and it doesn't increase the complexity of the sentence.

pgurley:
This looks good to me. The only other question I have, (which might be for @asabaylus ) is whether this should merge to master or release/3.0.

PR #226, opened 2016-08-25
[SYTO-734] Update button styles

hkekeli:
https://jira.natgeo.com/browse/SYTO-734

git checkout 
gulp serve
View that buttons styles are updated here: http://localhost:3000/patterns/


pgurley:
Looks good to me, merging

PR #229, opened 2016-08-26
[bugfix/SYTO-756] add a check for the presence of tracking codes

bpearsoncse:
This fixes a bug relating to https://jira.natgeo.com/browse/SYTO-756 in which undesired values (undefined, null) were being appended to the link URL if tracking codes were not present. This fixes that by adding an additional check and just passing the URL if codes are not present.


benjachapman:
LGTM

PR #240, opened 2016-08-30
[SYTO-770] Parallax Behavior Demo

benjachapman:
https://jira.natgeo.com/browse/SYTO-770

Demo page provided at /components/parallaxDemo

doesn't need to be merged, necessarily, but will serve as jumping off point for next parallax work


app/modules/parallaxDemo/docs.html
```diff
@@ -0,0 +1,127 @@
+<div class="pagewrapper mt3_row mt3_col-gut-sm mt3_col-md-gut-md mt3_col-lg-gut-lg">
```
pmourelle: All this was autom. generated? Is needed?

app/site/components/parallaxDemo/index.html
```diff
@@ -0,0 +1,11 @@
+---
+title: Parallax Demo
+rootPath: ../
+---
+<style>
+.headerwrapper {
+  display:none;
```
pmourelle: Why is this needed?

benjachapman: That css rule was to prevent the sticky header from appearing on this demo page, as the parallax image behavior assumes it is visible across the entire viewport (which will be the case on the homepage, generally)

benjachapman:
This work's functionality has been abstracted from the hard-coded example in this branch and incorporated into an actual component with this PR here: https://github.com/natgeo/mortar/pull/245. Therefore, i'll close this PR, its proof-of-concept purpose fulfilled

PR #245, opened 2016-09-06
[SYTO-754] Add Parallax effect to POD Component

pgurley:
https://jira.natgeo.com/browse/SYTO-754

To test:

- Pull SYTO-754
- Run `gulp serve`
- Navigate to http://localhost:3000/components/podpromo/index.html
- Verify parallax behavior at various break points


pmourelle:
Shouldn't this PR be combined with #240 ?

benjachapman:
@pmourelle I don't reckon so - #240 is more of a proof of concept than anything else. This PR abstracts that work into an actual usecase, so i think this one should win out at the moment.

benjachapman:
This behaves well locally, code looks good. We will likely refine our parallax behavior over time with UX around any performance concerns, but this is working solidly now. All tests passing.

PR #246, opened 2016-09-06
[feature/SYTO-793-2] design changes per Kevin

bpearsoncse:
These are design changes based on Kevin's comments in https://jira.natgeo.com/browse/SYTO-793


pgurley:
Merged

PR #252, opened 2016-09-07
switch to fluid gutters

xiwcx:
https://jira.natgeo.com/browse/SYTO-740

* make sure new gutter value is correct
* make sure all references to old gutters are purged
* make documentation is updated to reflect new gutter pattern


benjachapman:
LGTM - this seems to behave as advertised

PR #253, opened 2016-09-08
[SYTO-822] Create and Style The Latest Component

pgurley:
https://jira.natgeo.com/browse/SYTO-822

This story creates "The Latest" component, which is rendered by a component-data modeled (https://jira.natgeo.com/browse/AEM-5800) as "ContentPackage". The "ContentPackage" component will also contain the "Most Read" component. That work is slated for a future sprint. I imagine during the work for https://jira.natgeo.com/browse/SYTO-885 and https://jira.natgeo.com/browse/SYTO-886 there will be some refactoring of the instantiation of these components.

It should also be noted that the styling of the individual story cards is being handled in different stories as well.

To test:

- Pull feature/SYTO-822
- Run `npm test` to verify no errors
- Run `gulp serve`
- Load http://localhost:3000/contentPackages/contentPackage/
- Verify 4 cards are rendered
- Verify heading value can be updated
- Don't worry about layout of story cards yet


app/contentPackages/contentpackage/_styles.scss
```diff
@@ -0,0 +1,73 @@
+$mt3_black: #000;
+$mt3_white: #fff;
+$mt3_yellow: #ffcc00;
+$mt3_dark: #222222;
```
benjachapman: Are there actual Mortar color variables that could be used here? I think there is at least for "black"

app/contentPackages/contentpackage/scripts/components/FourUp.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+    let currentCard = card;
+    switch(index) {
+      case 1:
+        currentCard.config.aspectRatio = "16:9";
+      break;
+      case 0:
+        currentCard.config.aspectRatio = "3:2";
+      break;
+      default:
```
benjachapman: ought this "default" statement be removed? not sure if having a blank default case has some value i'm unaware of

app/contentPackages/contentpackage/_styles.scss
pmourelle: Actually, all of them :)
`/app/styles/themes/default/vars/_colors.scss`
`/app/styles/global/vars/_theme.scss`

app/contentPackages/contentpackage/scripts/ContentPackagePestle.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+import React from 'react';
+import * as ReactDOM from 'react-dom';
```
pmourelle: Is that `*` needed?

benjachapman:
Only a small nit or two to pick, looks great. tests & configurator working well.

The only concern i have here is the naming of the component, but I think we can address that on the go-forward. We'll need to discuss that with the larger group i reckon.

test/app/contentPackages/contentPackage/components/fourUp_test.js
```diff
@@ -0,0 +1,491 @@
+'use strict';
+
+import FourUpComponent from '../../../../../app/contentPackages/contentpackage/scripts/components/FourUp.jsx';
+import {shallow} from 'enzyme';
+import React from 'react';
+
+describe('FourUp Component', () => {
+  let el;
+  const options = {
+      'id': '3t35sdfj',
```
pmourelle: Would be much easier to read if all json lives in a separate file.

app/contentPackages/contentpackage/_styles.scss
pgurley: You caught my TODO I forgot about! Making this update

app/contentPackages/contentpackage/scripts/components/FourUp.jsx
pgurley: I propose adding a comment to the default value explaining what's going on vs. leaving it out altogether

Something like:
```
      default:
        // Default will be whatever was passed in as a prop. We're only overriding the first two cards
```

Thoughts?

app/contentPackages/contentpackage/scripts/ContentPackagePestle.js
pgurley: I don't believe so. Removing it

app/contentPackages/contentpackage/scripts/components/FourUp.jsx
benjachapman: ah, yeah fine by me

test/app/contentPackages/contentPackage/components/fourUp_test.js
pgurley: OK, I broke this out into a separate file

pgurley:
All feedback has been addressed and this is ready for another round

benjachapman:
Looks great, thanks Phil. merging.

pmourelle:
NOOICE!

PR #256, opened 2016-09-08
[SYTO-793-3] fix aspect ratio bug on mobile, set min-width

bpearsoncse:
Changes per comments in https://jira.natgeo.com/browse/SYTO-793


pgurley:
Looks good to me, merged!

PR #259, opened 2016-09-09
[WIP] Merging master into release/3.0

pmourelle:
This PR's intention is to integrate into mortar3 all work done over master branch (mortar2).
This is still a WIP as it needs some more checking as `mt2` vs `mt3` classes should be discussed and resolved (if we translate all of them to mt3 or if both will co-exist).


sebasrgl:
We are going to create a new Pull Request to handle the merge between release/3.0 and master

PR #263, opened 2016-09-09
[feature/SYTO-819-2] Style the Story Card -- Video

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-819

To Test:
pull feature/SYTO-819-2 (I created a new branch because of all the merge conflict craziness earlier)
```
gulp serve
```
verify all tests pass and matches AC


pmourelle:
image
Texts under promos are too tiny, 6px for kickers and title and description seem also too small.

benjachapman:
Like Pedro, i have a few styling concerns, though I'm not sure this is the ticket that would address them - how are promos expected to pad text for mobile? They are currently full-bleed with both image and text, going against the designs: 

screen shot 2016-09-12 at 9 27 57 am

That said, i'm not sure how that's supposed to be achieved, if the component is always reading its own width rather than the width of the page - in theory it would be a similar width for mobile as it is at larger breakpoints.

I'm also concerned that the video thumbnail button does not have a maximum-width, and can get to what i find to be too large of a size at large breakpoints.

However, this work doesn't encompass those items, so we should bring them up and address them with regards to QAing promo cards as a whole I believe - merging this to get these small updates in

PR #266, opened 2016-09-12
[SYTO-822] Small tweak to letter-spacing per feedback

pgurley:
To test:
Pull SYTO-822-3
Run `gulp serve`
Verify `letter-spacing` on header element with the className `mt3_fourup-header` is `.1em` on content package component at http://localhost:3000/contentPackages/

This change also fixes linting errors on app/modules/videoplaylist/_styles.scss, so layout on the playlist should be confirmed as well


arieldelfor:
Lookgs great to me. Merging it.
Tests passing ok :+1: 
image

PR #267, opened 2016-09-12
[feature/SYTO-808-2] design changes per Kevin

bpearsoncse:
These are changes based on Kevin's feedback in https://jira.natgeo.com/browse/SYTO-808 as-well-as incorporating the button hover state to include the entire card area per his comment in https://jira.natgeo.com/browse/SYTO-819


app/modules/promocard/_styles.scss
```diff
@@ -103,11 +102,11 @@ $mt3_gradientIndex: 3;
   .mt3_promocard-kicker {
     @extend .mt3_kicker;
     float: left;
-    font-size: 0.25rem;
+    font-size: 0.75rem;
```
pgurley: I believe the font-size here is repetitive and unnecessary since you're extending .mt3_kicker

app/modules/promocard/_styles.scss
```diff
@@ -103,11 +102,11 @@ $mt3_gradientIndex: 3;
...
     line-height: 1;
     margin-bottom: 10px;
     position: relative;
-    text-decoration: none;
+    text-decoration: underline;
```
pgurley: If I'm reading the feedback correctly, the underline should only be on hover

pgurley:
The kicker is overlaying the Play button now. Is that in scope for this review? 
screen shot 2016-09-12 at 11 17 51 pm

pgurley:
I believe all feedback has been addressed. Merging

PR #276, opened 2016-09-16
[AEM-5957] Fix a few odds and ends related to using m3 components in AEM

benjachapman:
in the work for https://jira.natgeo.com/browse/AEM-5957, we identified a couple small changes needed in Mortar - 

1. react-element-query is only a devDependency - it should be a peerDependency
2. the contents of the app/scripts/util file should be available in /lib
3. small tweaks to the Portal components

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@natgeo/mortar_v3",
-  "version": "1.0.7",
+  "version": "1.0.8",
   "description": "a living styleguide for National Geographic Partners",
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "jquery": "2.x",
     "react": "^0.14.8 || ^15.1.0",
     "react-dom": "^0.14.8 || ^15.1.0",
+    "react-element-query": "^2.0.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "jquery": "2.x",
     "react": "^0.14.8 || ^15.1.0",
     "react-dom": "^0.14.8 || ^15.1.0",
+    "react-element-query": "^2.0.2",
```
arieldelfor: I agree with this but you should not lock your peer dependencies down to specific patch versions.

Also i am not sure if you still need to add this module at dependencies. Otherwise, a fresh install of Mortar 3 would throw an error like `'npm ERR! peerinvalid The package does not satisfy its peerDependencies requirements!'`

Cheers.

asabaylus: Peer deps should be as loose as possible

benjachapman: Should we make it a dependency, then? This would not be part of AEM otherwise - this is m3 specific thus far

arieldelfor: Both things, make it a dependency and peer-dependency.
You are free to specify a patch version under dependencies. Don't specify patch version at peer-dependencies. :)

krlospelaez:
PR Ok

PR #278, opened 2016-09-16
[SYTO-897] Create border classnames for social colors

hkekeli:
https://jira.natgeo.com/browse/SYTO-897

On the color page, use the console to verify that the border class names specified in the AC work.


pmourelle:
LGTM

PR #279, opened 2016-09-16
[SYTO-912] Update column gutter documentation

hkekeli:
https://jira.natgeo.com/browse/SYTO-912

Instructions:

On the layout page, look for copy updates and the new column gutter documentation specified in the AC.


sebasrgl:
Looks good :shipit:

PR #280, opened 2016-09-16
[SYTO-830] Mortar refactor to improve build process and modules exporting

krlospelaez:
# Atention: next code contains important changes on Mortar repo

- "app" folder renamed to "src"
- All jsx files where renamed to js, to provide a much more consistent building process.
- The files contained on "scripts" folder on each module or contentPackage were moved one folder back and  the "scripts" were removed. This folder was useless and it created much more boilerplate when importing the modules
- Pestle has now a consistent name to prevent more confusions about how to import it. Now is "@natgeo/pestle" instead of "@natgeo/mortar-pestle"

To test:

- gulp serve
- Verify modules are working as expected


pmourelle:
Works fine!

benjachapman:
Hi @krlospelaez - I noticed JS modules no longer being compiled to /lib on build, with these changes. I may be missing how you're intending those to be consumed - are they to be directly referenced in /src?

benjachapman:
Also, i'm seeing this work is causing some linting errors, perhaps owing to the files in question being "js" rather than "jsx" now?

krlospelaez:
Hi @benjachapman, it's being compiled on /lib, you can confirm it running `gulp build`.

About the eslint, these eslint errors about `strict mode` are going to be removed on this ticket https://jira.natgeo.com/browse/SYTO-758, what's why I didn't fix them on this ticket and yeah these errors are caused by the extension change from jsx to js but it's not because I introduced them, they had been there all this time, it's just eslint was not checking jsx extension.

PR #283, opened 2016-09-20
[SYTO-893] FE: Handle manual crops in promo cards

krlospelaez:
To test:

- gulp serve
- Verify first Promo Card on component menu is using cropping srcsets.


benjachapman:
LGTM, i added a few tweaks as well

PR #282, opened 2016-09-20
[SYTO-930] Hero  W/ Two Rails styling tweaks

benjachapman:
https://jira.natgeo.com/browse/SYTO-930

This tightens up the styling of the HwTR, as well as a few assorted issues with the promo card

PR #284, opened 2016-09-20
[SYTO-891] HWTR - Allow the component heading to go above and below the hero

benjachapman:
https://jira.natgeo.com/browse/SYTO-891

Note that this work builds off the work in SYTO-930, currently in PR (it should be merged first, ideally): https://github.com/natgeo/mortar/pull/282

As Advertised! Note that this also disabled parallax on the HwTR for now, per the ticket - i did enable it to fire, however, when a query var of "enableParallax" is provided on the URL


pgurley:
Merged

PR #286, opened 2016-09-20
[SYTO-925] add broadsheet content

xiwcx:
https://jira.natgeo.com/browse/SYTO-925

* look at rendered content
* make sure it matches AC


krlospelaez:
PR Ok

PR #291, opened 2016-09-21
[SYTO-856] - Cannot navigate to videos when pre-roll is playing.

arieldelfor:
A fix for when the playlist intends to play another item while pre-roll (bumper) is running.

Notes:
 - A small change for Video was made, Video component will not re-rendering unless the props have changed.
 - modules-video bumped to 1.2.6

- git checkout feature/SYTO-856
- gulp serve
- scroll down to playlist, play a video and when pre-roll is shown immediately click on other video from the playlist. The player must play the new pre-roll thus play the selected video.

----
https://jira.natgeo.com/browse/SYTO-856


pmourelle:
LGTM!

PR #293, opened 2016-09-21
[SYTO-920] Fix aspect ratio of Video Play button overlay

hkekeli:
https://jira.natgeo.com/browse/SYTO-920

To test on the UI Patterns and component pages --

Confirm that the video Play button overlay now looks like this:

fix-triangle

to reflect the intended design here: 
intended-design


pgurley:
Merged

PR #295, opened 2016-09-22
[AEM-5956] Bumping version number for ingestion into AEM

pgurley:
This change bumps the version number up to get latest code into AEM


benjachapman:
yep!

PR #297, opened 2016-09-22
[SYTO-878] - Add duration and kickers to Video Playlist component.

arieldelfor:
This PR adds Kicker duration and kicker label to thumbnails.

 - git checkout feature/SYTO-878
 - gulp serve
 - localhost:3000/components/videoplaylist/


src/modules/videoplaylist/VideoCaption.js
```diff
@@ -50,9 +50,13 @@ class VideoCaption extends Component {
   }
 
   render() {
-    const {title, abstract} = this.props;
+    const { duration, kickerLabel, title, abstract} = this.props;
     return (
       <div className="mt3_video-playlist--current-information mt3_bgcolor--gray80">
+        <div className="mt3_kicker-wrapper">
+          {(kickerLabel) ? <span className="mt3_kicker mt3_color--white">{kickerLabel}</span> : ''}
```
pmourelle: If kickerLabel and duration aren't provided (as below), both `span`s are created and left empty.
```
duration: undefined
kickerLabel: ""
```

src/modules/videoplaylist/VideoThumbnail.js
```diff
@@ -4,6 +4,17 @@ import React, {Component} from 'react';
 import Image from '@natgeo/modules-images';
 import Truncate from 'react-truncate';
 
+export const KICKER_TYPES = ['now', 'next'];
+
+const Kicker = ({type}) => (
+  <div className="mt3_kicker-wrapper">
+    { (type === 'now')
+      ? <span className="mt3_kicker mt3_color--black kicker-now">now playing</span>
+      : <span className="mt3_kicker mt3_color--primary kicker-next">next</span>
```
pmourelle: What about being explicit on checking that `type === 'next'` in case a new type is added?

src/modules/videoplaylist/VideoCaption.js
jaanauati: yeah, react wraps '' with ```span```, if you want to avoid this i think you should fallback to null, example:
```
{ kickerLabel ? <span className="mt3_kicker mt3_color--white">{kickerLabel}</span> : null }
```

src/modules/videoplaylist/VideoThumbnail.js
arieldelfor: @pmourelle because it is explicit at prop requirements (kickerType: React.PropTypes.oneOf(KICKER_TYPES))
If at some case someone upgrades this component, then it means there will be an upgrade at the render logic as well thus another conditional will be added.
It is deterministic already.
Do you think it should be modified anyway?

src/modules/videoplaylist/VideoThumbnail.js
```diff
@@ -24,6 +36,7 @@ class VideoThumbnail extends Component {
       <div className={this.props.wrapperClass}>
         <Image {...imageModel} />
         <a href={this.props.item.path} className="thumbnail-overlay mt3_none" title={this.props.item.title} data-guid={this.props.item.guid} onClick={this.onClick.bind(this)}>
+          { kickerType && kickerType.length && <Kicker type={kickerType}/>}
```
arieldelfor: I thought this way was more readable.
I am taking it back to your suggestion.

src/modules/videoplaylist/VideoCaption.js
arieldelfor: i could simply use `kickerLabel && <...>` without take care of null.

Anyway i have to fix it.
Thanks for reviewing.

src/modules/videoplaylist/VideoThumbnail.js
pmourelle: PropTypes defines validity of the props but in rendering you're assuming you'll get "the other" value. Well, I'd rather to be over explicit than imply the `else`, but it's up to you **goes away slowly with pity eyes**

src/modules/videoplaylist/VideoThumbnail.js
pmourelle: If you could ensure that kickerType will never be `""` then you may save the length check and stick with your current [simplified] line.
`kickerType  && <Kicker type={kickerType}/>`

src/modules/videoplaylist/_styles.scss
```diff
@@ -206,6 +206,17 @@
       }
     }
 
+    .mt3_kicker-wrapper {
+      margin-bottom: .5rem;
+
+      .mt3_color--black {
```
pmourelle: This should be `kicker-now` and below `kicker-next`.

pmourelle:
LGTM!

PR #299, opened 2016-09-22
[SYTO-916] Playlist: Video does not load on FF 

sebasrgl:
https://jira.natgeo.com/browse/SYTO-916

Just for check on FF and Edge.
The implemented approach was different in order to keep the layout consistency among the browsers with the video description and the video.

#### How to test
1. ``gulp serve``
2. Open /components
3. Scroll down until the video playlist section
4. The video into the VideoPlaylist should load properly with aspect ratio 16:9.


src/modules/video/video.js
```diff
@@ -70,15 +70,16 @@ class Video extends Component {
 
     if (lazyLoad === true) {
       return (
-        <figure itemType="http://schema.org/VideoObject" className={`mt3_video mt3_videopromo-container mt3_intratio--broadcast mt3_bgcolor--black ${className}` }>
-          <LazyLoad offsetVertical={200} onContentVisible={this.handleOnLoad.bind(this)}><span>&nbsp;</span></LazyLoad>
-          <div className="mt3_video-wrapper" dangerouslySetInnerHTML={{__html: this.videoContainer}}>
-          </div>
+        <figure itemType="http://schema.org/VideoObject" className={`mt3_video mt3_videopromo-container mt3_bgcolor--black ${className}` }>
+          <LazyLoad offsetVertical={200} onContentVisible={this.handleOnLoad.bind(this)}>
```
arieldelfor: @sebasrgl Could you give a brief explanation why are you moving this into lazyload block?
At AEM there was a special fix for this, where it should be outside of lazyload in order to prevent side effects. I don't remember very well but i think it was @benjachapman who discovered this suggestion.

sebasrgl: @arieldelfor That was moved there because i was getting a gap between the gray container(the div with the video title and the description) and the video player, the problem was the span inside the LazyLoad element. I removed the span but i never got the video, just the placeholder.
So, I had to move there for that reason and the video load properly with this configuration.

benjachapman: @arieldelfor, indeed, that was one of the attempted fixes we added to AEM hoping to help the "gray box" issue. We did that because we theorized that a potential problem could be caused by modules-video not being invoked until the LazyLoaded content was fired. I don't believe we really know whether or not that helped, or to what degree - it certainly did not fix the issue. 

The change in AEM also made sense to me because the only thing we really needed to happen when the LazyLoad element fired was for the calback function to fire. If it's the same to you @sebasrgl, i'd only advise doing this in such a way so that a placeholder (which is currently styled on the "mt_3_video-wrapper" element) displays before the LazyLoad handler has fired.

arieldelfor:
Thanks for taking care on the advice.
looks great to me.

PR #302, opened 2016-09-23
[SYTO-879] - Update playlist to handle 5+ videos.

arieldelfor:
JIRA: https://jira.natgeo.com/browse/SYTO-879
----

- git checkout feature/SYTO-879
- gulp serve
- videoplaylist's thumbnails should be scroll-able only at desktop.
- it should have a black gradient at bottom.


sebasrgl:
Looks good :shipit:

PR #306, opened 2016-09-23
[SYTO-885] Don't change the color of the headline on hover afterall

pgurley:
To Test

- Pull SYTO-885-3
- Run `npm test` to verify tests/linting
- Run `gulp serve`
- Navigate over to http://localhost:3000/contentPackages/ and verify that the titles remain white when a card is hovered over


benjachapman:
👍

PR #308, opened 2016-09-26
[SYTO-969] override user agent default margin in Safari

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-969


benjachapman:
Looks good. I think this speaks to our issue of not having a real CSS reset, but this will fix this particular display error jus' fine

PR #309, opened 2016-09-26
[SYTO-892] (second one) Correct Display error in HWTR when there's no hero promo

benjachapman:
https://jira.natgeo.com/browse/SYTO-892

this corrects the two display errors Kevin mentions above

to test - checkout SYTO-892-2
- confirm mentioned display errors do not occur


bpearsoncse:
LGTM 👍

PR #311, opened 2016-09-28
[SYTO-856] - Fix videoplaylist 'selected item not available' message

arieldelfor:
Sometimes a message is shown 'Current selected item is not currently availabe' when attemping to load a video from playlist.
This PR intends to fix this.

- git checkout feature/SYTO-856-2
- gulp serve
- navigate and paly videos from palylist.


pmourelle:
Like a charm.

PR #314, opened 2016-09-29
[broadsheet-bugfix] fix component error in broadsheet

bpearsoncse:
This fixes the Uncaught Reference Error produced by broadsheet. There is another JS error still, but it does not affect page rendering


benjachapman:
👍

PR #317, opened 2016-09-30
[SYTO-987] Modify Text Only display for Story Card Component

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-987


src/modules/promocard/components/shared/PromoText.js
```diff
@@ -1,81 +1,149 @@
...
+                {(props.type === 'video' && props.text.duration && props.text.kicker.style !== 'prompt') ?
+                  <div className={`${subheadColor} ${attrs.className} mt3_card-subhead--right`}>{props.text.duration}</div>
+                  : null
+                }
+                {(props.type === 'gallery' && props.text.photoCount && props.text.kicker.style !== 'prompt') ?
+                  <div className={`${subheadColor} ${attrs.className} mt3_card-subhead--right`}>{props.text.photoCount} Photos</div>
+                  : null
+                }
+              </div>
+            </div> : null }
```
benjachapman: does this section of kicker logic get output identically in these different sections? May be best to define it as a variable, if so, to avoid boilerplate

bpearsoncse: agreed, that should probably be put into a variable. layout of the kicker changes in relation to the title and dek, but the kicker itself does not change.

benjachapman:
Nice work, merging with a slight CSS tweak

PR #319, opened 2016-09-30
[SYTO-924] Create pattern for Channel Mapping

mwise-cse:
* Added Channel Branding to Patterns
* Updated Show Card to use Channel Branding Pattern
* Updated Show Card FE test for correct Channel Branding theme class

https://jira.natgeo.com/browse/SYTO-924


bpearsoncse:
👍  Works as advertised

PR #322, opened 2016-10-04
[SYTO-907] Edit code snippet in the POD Configurator

hkekeli:
https://jira.natgeo.com/browse/SYTO-907

Update: Added prism.js

**Instructions:** 
-npm install 
-Configure options on this page http://localhost:3000/components/podpromo/index.html and make sure: 

a) The configurator component works as expected
b) the html part of the code snippets retain their syntax highlighting. 

**Bonus**: JS code snippets are now using syntax hightlighting. Hooray.

**Please note**: Adding indentation to the snippets is going to be difficult because this stuff isn’t written by hand.


src/site/_styles/docs.scss
```diff
@@ -420,6 +422,19 @@ pre {
   }
 }
 
+.configurator-markup,
+.configurator-codesnippet {
+  //Styles imported from prism to accomodate the configurator codesnippet
+  font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
```
pmourelle: `font-size` wouldn't have to be defined also? Browsers tend to do whatever they want with `pre` elements. And as this seems like to be used a lot for whenever we want to put some coding, we can have it defined in some variables?

src/contentPackages/configurator/CodeSnippet.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 import React, {Component, PropTypes} from 'react';
+import * as Prism from 'prismjs';
```
sebasrgl: This library is missing on package.json

sebasrgl: In order to avoid this use the ``-S`` flag and that way you keep the dependency on the package.json e.g. ``npm install -S prismjs``

src/site/_styles/docs.scss
hkekeli: Updated!

src/contentPackages/configurator/CodeSnippet.js
hkekeli: Updated!

pmourelle:
LGTM

PR #323, opened 2016-10-04
[AEM-5959] Parallax Refactoring

benjachapman:
https://jira.natgeo.com/browse/AEM-5959

This work refactors the logic for Parallax positioning in Mortar. It removes the use of GSAP/ScrollMagic, opting for a more customized implementation. Decisions made were performance-based, and the three major things that have brought improvements are:

1. Animated elements become "position:fixed" while being parallaxed (they had been "absolute" previously). This let to many modifications to the measurements taken/required, and complicated the resize process, but this change greatly smooths over the "jittery" effects we'd seen previously, especially in Safari/IE
2. Rather than basing updates on scroll events, when there is at least 1 parallax scene on the page, an interval will be called (currently every 10 milliseconds) that smartly updates the scenes currently in the viewport. In addition, each update first calls "window.requestAnimationFrame" to ensure the page is ready to repaint.
3. Not transitioning/animating: Because of how incrementally the parallax effects are applied, and the fact that we are not animating the size or scale of any elements (only their y position) it actually turns out to be more reliable to not animate the tween when new pixel values are applied to parallax elements. With the 10ms interval time, these effects typically occur at a higher Framerate than the previous implementation with GSAP/ScrollMagic, with the difference most notable in Safari/IE

ScrollMagic is no longer in use. GSAP has no current usecase either, but one potentially might be the Broadsheet, when it is done - the library can be re-added then if needed.

Unit Tests coming for the "parallaxSceneManager" module, but I want to get this in front of design to make a ruling on the performance improvements. Things are much more solid from my perspective, but I want to be sure


benjachapman:
update: added p'lax behavior for MostRead/Latest

pgurley:
Appears to function well and existing tests and linting all pass. I think it's good to get design eyes on behavior

PR #325, opened 2016-10-05
[AEM-6200] TEST- HwTR and Manual Crops in AEM 

mwise-cse:
Added fix for manual crops

https://jira.natgeo.com/browse/AEM-6200


benjachapman:
👍

PR #328, opened 2016-10-06
[SYTO-1010] Show Card text not vertically centered in Safari

mwise-cse:
Added div and updated sytles to fix show card centering in safari

https://jira.natgeo.com/browse/SYTO-1010


benjachapman:
Looks good, works for me

PR #329, opened 2016-10-06
Cleanup

krlospelaez:


benjachapman:
This test fine locally for me! I reckon we should keep an eye on this in AEM, and see if corresponding changes are necessary?

PR #334, opened 2016-10-07
[AEM-5958] HwTR parallax behavior for when heading is "above" hero

benjachapman:
https://jira.natgeo.com/browse/AEM-5958

This also fixes some bugs with the MR/L and the PodPromo not canceling parallax when configured to do so

Behavior has been walked-thru with Kev

The "above" the hero configuration can be seen on the /contentPackages page (not the configurator)


bpearsoncse:
Smoove as butta. :shipit:

PR #335, opened 2016-10-07
[SYTO-1007]: remove body padding, plus some assorted bug fixes

benjachapman:
https://jira.natgeo.com/browse/SYTO-1007

as advertised

other bug fixes:
- removed odd spacing above MR/L
- promo text displays when there's not a kicker
- images render when a url is passed but not a srcset
- POD background image no longer obscures neighboring elements


bpearsoncse:
👍  LGTM

PR #336, opened 2016-10-10
[SYTO-887] Clicking Video Card Opens the Video Modal

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-887

Quite a bit of refactoring done to this component to use Pub/Sub, decoupling it from the component it is being called from and to allow for only one instance of the component being necessary on the page. This borrows heavily from the VideoModal component in AEM, in the hopes to further ease implementation on that platform as well. Additionally, this refactor resolves the issue of the video continuing to play after the modal is closed.


benjachapman:
LGTM!

PR #341, opened 2016-10-12
[AEM-6339] Fixes bug where kicker doesn't render if it's empty in AEM

pgurley:
https://jira.natgeo.com/browse/AEM-6399

To test:

- Pull AEM-6399
- Run `npm test` to verify tests/linting
- Run `gulp serve`, verify component renders
- Run `gulp build`
- From within the mortar directory, `cp lib/modules/podpromo/PodPromo.js ~/Documents/workspace/ngs/aem-all/aem-platform/platform-view/node_modules/@natgeo/mortar_v3/lib/modules/podpromo/.` (**note second path should match wherever you have your AEM code)**
- From within the mortar directory, `cp lib/modules/podpromo/_styles.scss ~/Documents/workspace/ngs/aem-all/aem-platform/platform-view/node_modules/@natgeo/mortar_v3/lib/modules/podpromo/.` (**note second path should match wherever you have your AEM code)**
- From platform-view in AEM, run `mvn clean install -Pauto-deploy -s ../settings.xml`
- Create POD component with a kicker, verify it renders.
- Remove the kicker and verify the POD component still renders
- Hit the merge button


src/modules/podpromo/configurator/MTPodPromoConfiguratorPestle.js
```diff
@@ -46,10 +46,7 @@ class PodPromoConfigurator extends Module {
           affiliation: 'National Geographic',
           assetSource: '',
           kicker: {
-            label: 'Kicker Label',
-            url: '/content/news/en_US/weird-wild.html',
-            target: '_self',
-            trackingCodes: '?utm_medium=site&utm_source=ng.com'
+            label: '<a href="http://www.google.com">Kicker Label</a>'
```
benjachapman: Hmm, i worry that links might be passed to the kicker through the old method as well (by passing url, target and trackingCodes values). We may need to attempt to support both - let's ask in planning

pgurley: screen shot 2016-10-12 at 8 30 43 am

Yea, the original data model indicated it'd come in as separate fields, but this is what AEM is currently passing in via a rich text field

bpearsoncse: Maybe there could be a BE change for that similar to what @parmstrong62387 did in https://jira.natgeo.com/browse/AEM-6193 (see comments). That way if it is an override from rich text field it gets a different prop and makes it really easy to check.

pgurley: Maybe. I'm not sure why a solitary component like this would need to receive data in two different formats though. We can figure it out in planning ¯\_(ツ)_/¯

pgurley: OK @benjachapman , I've grudgingly made this update. POD now supports both potential data models for a kicker. In the mortar components each is represented, one style on the main components page and the other on the POD configurator

benjachapman:
thanks Phil, looks good

PR #345, opened 2016-10-13
SYTO-1036: prevent HwTR's heading from going to two lines

benjachapman:
https://jira.natgeo.com/browse/SYTO-1036

As Advertised. check out the HwTR on the /contentPackages page for an example


bpearsoncse:
👍  works as advertised

PR #348, opened 2016-10-13
[SYTO-1015] First pass at handling rich text in pod and latest/most read components

pgurley:
https://jira.natgeo.com/browse/SYTO-1015

Updates for rich text compatibility on POD component as well as most read. Kicker logic was updated on most read to accommodate this.

To test:

- pull SYTO-1015
- npm test
- gulp serve
- verify fields are displaying properly on configurator pages as well as the components and content packages page.


benjachapman:
Looks good to me. I observe no negative consequences from the inline-block rule on the new mixin, but lets' keep an eye on that

pgurley:
@benjachapman Agreed about the inline-block rule. If it turns out to be too specific of a use case we can move it to the component's scss file

PR #349, opened 2016-10-13
[SYTO-1018] Add Margin to Show Card

mwise-cse:
Added bottom margin to show card

https://jira.natgeo.com/browse/SYTO-1018


benjachapman:
splendid

PR #352, opened 2016-10-13
[SYTO-979] Handle richtext for card components in Mortar

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-979

Also added to video modal


benjachapman:
I'm gonna add some "multiple paragraph" logic here, as I think video modals will need it

bpearsoncse:
Sounds good

src/modules/videoplaylist/VideoCaption.js
```diff
@@ -55,11 +55,8 @@ class VideoCaption extends Component {
     return (
       <div className="mt3_video-playlist--current-information mt3_bgcolor--gray80">
         <div className="mt3_kicker-wrapper">
-          {(kicker && kicker.label) ? <a className="mt3_kicker mt3_color--white" href={kicker.url} target={kicker.target}>{kicker.label}</a>
-            : (kicker && kicker.html) ? <span className="mt3_kicker mt3_color--white" dangerouslySetInnerHTML={{__html: kicker.html}} />
-            : null
-          }
-          {(duration) ? <span className="mt3_kicker mt3_color--white">{duration}</span> : null}
+          {(kicker && kicker.label) ? <a className="mt3_kicker mt3_color--white" href={kicker.url} target={kicker.target} dangerouslySetInnerHTML={{__html: kicker.label}} /> : null}
```
pgurley: @bpearsoncse It looks like this will put <a/> tag inside the <a/> tag if the label comes in rich text
screen shot 2016-10-14 at 10 08 41 am

benjachapman:
I added support for multiple <p> tags and added an example to the Video card example in the Promo Card example. Also, I added a change that made the decision for whether or not the <p> tag is set to inline-block based on an argument passed to the mixin. Ben, can you check that against your work?

src/modules/videoplaylist/VideoCaption.js
bpearsoncse: Good catch. I'll bring over the logic from the PromoText component.

bpearsoncse:
Yeah I'll check it when I make Phil's change

benjachapman:
@pgurley can you also double check that my mixin changes didn't björk any of your stuff?

src/modules/videomodal/_styles.scss
```diff
@@ -1,6 +1,39 @@
...
+    text-decoration: none;
+    z-index: $mt3_z-middle + 10;
+
+    a:hover, a.mt3_promocard-kicker:hover {
+      cursor: pointer;
+      text-decoration: underline;
+    }
+
+    &--inverse {
+      border-right: 1px solid white;
```
pmourelle: I think an `inverse` color variable would be needed, so instead of `white` we may use a `mt3_color-inverse (being set to $mt3_color-white)` for example.

src/modules/videoplaylist/VideoCaption.js
pmourelle: So, just kicker data in json should be changed ? Or is anything else to take care of in here?

bpearsoncse:
Changes made - back to you @pgurley @benjachapman

benjachapman:
👍

PR #354, opened 2016-10-14
[AEM-6401] Text-Only Card has wrong display in dark theme

mwise-cse:
* Updated text only promo styles 
* Added text only promo to htr content package

https://jira.natgeo.com/browse/AEM-6401


src/modules/promocard/components/shared/PromoText.js
```diff
@@ -16,7 +16,12 @@ class PromoText extends Component {
     const overlayClasses = props.config.overlay ? 'mt3_color--white mt3_promocard-nested-text' : 'mt3_color--neutral--xxd';
     const subheadColor = props.config.overlay ? 'mt3_color--white' : 'mt3_color--gray40';
     const inverseTitle = props.config.overlay || props.theme === 'dark' || props.noImages ? 'mt3_color--white' : 'mt3_color--black';
-    const inverseDek = props.theme === 'dark' ? 'mt3_color--white mt3_promocard-dek--inverse' : props.noImages ? 'mt3_color--white mt3_promocard-dek--text-only' : 'mt3_color--black';
+    
+    const inverseDek = props.theme === 'dark' & !props.noImages ? 'mt3_color--white mt3_promocard-dek--inverse'
```
pmourelle: I think `classnames()` may be useful here to make these lines more readable.

src/modules/promocard/components/types/Story.js
```diff
@@ -203,7 +204,7 @@ class Story extends Component {
     }
 
     return (
-      <div className={`mt3_row mt3_col-12 mt3_promocard-container ${bkgColor} ${kickerStyle} ${additionalClasses}`}
+      <div className={`mt3_row mt3_col-12 mt3_promocard-container ${bkgColor} ${noImagesClass} ${kickerStyle} ${additionalClasses}`}
```
alexrqs: as @pmourelle point this a very good use case for classname repo something like
```js
const promoContainerClass = cx({
  'mt3_promocard-container--text-only': noImages,
  .
  .
  .
});

<... className={promoContainerClass} />
```

this recommendation is because https://facebook.github.io/react/docs/class-name-manipulation.html
more examples https://github.com/JedWatson/classnames#usage

here is the current use of cx inside GN4 as friendly example
https://github.com/natgeo/modules-global-nav/blob/master/src/globalnavigation/regiondropdown/RegionDropdown.js#L56-L61

src/modules/promocard/components/shared/PromoText.js
```diff
@@ -16,7 +16,12 @@ class PromoText extends Component {
...
+      : props.theme === 'dark' && props.noImages ? 'mt3_promocard-dek--text-only'
+      : props.noImages ? 'mt3_color--white mt3_promocard-dek--text-only'
```
alexrqs: also props is it's being used too many times, destructuring pattern is a good option here saving some keystrokes and improving readability.

e.g. https://github.com/natgeo/modules-global-nav/blob/master/src/globalnavigation/expandedglobalnavigation/ExpandedGlobalNavigation.js#L12-L27

benjachapman:
Agreed on all points @alexrqs and @pmourelle, though I think this small ticket is perhaps not the place for such syntax changes. I think such refinement will come once we get through the bug list

pmourelle:
@benjachapman indeed, we'd need to clean under the carpet, tackle some bug/cleanup tickets to cover all or almost-all the cases. Still, we should be comfortable with any code being edited/added from now on. So, let's stop growing the carpet mountain :).

alexrqs:
@benjachapman usually I'll be #pro_move_big_changes_to_a_different_ticket but, is a less than 5 minutes change I don't think there is any harm :) and step by step we get a good repo.

what you think @mwise-cse ?

mwise-cse:
@alexrqs I'm okay with making this change as it should be really small. Assuming @benjachapman agrees.

benjachapman:
by all means :) i reckon there are many places in our code at the moment that could benefit from the same treatment, so let's do that as we can while still prioritizing getting bug fixes in front of QA

alexrqs:
awesome :+1:

mwise-cse:
@benjachapman @alexrqs Please take a look at my updates. Any more refactoring should most likely be done under a different ticket.

bpearsoncse:
Agreed @mwise-cse. Tested locally to confirm functionality in AEM. 👍  LGTM

pmourelle:
Changes look really good guys. Far easier to read and mantain :). Thanks for spending some extra time on this 👍

PR #355, opened 2016-10-14

mwise-cse:
* Updated text only promo styles for large breakpoint
* Updated text only promo padding when in hero of htr


benjachapman:
passing back to update when the styles are applied (from "being in the Hero of the Hero With Two Rails" to "when the promo is super wide")

benjachapman:
LGTM

PR #356, opened 2016-10-14
[SYTO-1043] Refine how parallax recalculates measurements

benjachapman:
https://jira.natgeo.com/browse/SYTO-1043

This ticket refines how the parallax system re-calculates its measurements when needed. 

Previously, the only situation that would cause the system to re-initialize its calculations was a resize event (debounced of course). There was also a listener (called at a 3 second interval) that would check for changes in the height of the document, but if a change was found, it would only update the scrollTop positions of child scenes, not their fundamental measurements.

After this work, that same document height listener will fire the same behavior as a resize event should a change in the document's height be found. 

To test: 
- build dat
- test scenarios involving at least 1 of the 3 components that use parallax: POD, HwTR, and MR/L
- try messing with things by making a DOM element suddenly a new height, and make sure parallax system reacts after a moment


pgurley:
This appears to work as described. I may have stumbled on and edge case, but we can keep an eye on that during testing

PR #359, opened 2016-10-17
[SYTO-1011] Make Channel Video work in modules video and video playlist

parmstrong62387:
- Added account parameter to VideoPlaylist, Video Modal, and Story Card
- Account is set to be an optional parameter, since a default exists in Modules Video

To test:
- Deploy updates from this branch to your AEM instance
- Install content package from ticket (https://jira.natgeo.com/secure/attachment/90586/AEM-5200-channel-videos-uuid-removed.zip)
- Ensure that a story card which points directly to a channel video or a NG video will function
- Ensure a video playlist rebrand with channel video or NG video will function


bpearsoncse:
tested locally - 👍  LGTM!

PR #358, opened 2016-10-17
[AEM-6485] Fixes various small bugs with pod promo and kickers in most read for AEM-6486

pgurley:
https://jira.natgeo.com/browse/AEM-6485
https://jira.natgeo.com/browse/AEM-6486

- Fixes bad url when AEM doesn't provide tracking codes for POD Promo
- Fixes layering issue with CTA link on mobile for POD Promo
- Attempts to fix AEM rendering bug with pipes around kickers in Most Read component

To test:

- Pull AEM-6485
- Run `npm test`
- Run `gulp serve`
- Verify POD promo link
- Verify CTA on POD Promo is clickable at both desktop and mobile breakpoints
- Verify kickers have left and right pipes in Most Read component

PR #360, opened 2016-10-17
[AEM-6485] Fix for CTA when there's no link

pgurley:
Follow up to https://github.com/natgeo/mortar/pull/358

Fixes issue with CTA when there's no CTA link in the Data model

This can be tested by pulling AEM-6485-2

There should be a CTA on http://localhost:3000/components/podpromo/index.html
There should be no CTA on the POD promo on http://localhost:3000/components/


bpearsoncse:
👍

PR #363, opened 2016-10-18
[AEM-6444] POD Parallax behavior broken in firefox

mwise-cse:
Added a fix for pod promo overflow when using Firefox in AEM by adding a negative z-index to the pod promo image wrapper div and by ensuring the HTR always has a background. This fix does not protect against two adjacent pod promos.

Note: This should also fix the pod promo overflow in IE and Edge.

https://jira.natgeo.com/browse/AEM-6444


benjachapman:
👍  let's try this out in AEM

PR #364, opened 2016-10-18

mwise-cse:
Update text only promo to use mortar classes for width and padding for better alignment with the rails in HTR.

Addresses Kevin's comments in this ticket.


bpearsoncse:
👍

PR #365, opened 2016-10-18
[SYTO-1059] Update letter-spacing of Verlag and Haas

hkekeli:
https://jira.natgeo.com/browse/SYTO-1059

Confirm the letter-spacing of both fonts are updated according to the AC.


pmourelle:
Looks good!

PR #368, opened 2016-10-19
[SYTO-1058] Story Card Design Tweaks

mwise-cse:
https://jira.natgeo.com/browse/SYTO-1058

Also, includes update to text only promo when in the hero position of HTR.
https://jira.natgeo.com/browse/SYTO-1074


bpearsoncse:
Tested locally - works as advertised :shipit:

PR #369, opened 2016-10-19
[SYTO-1056] Fixing layout issues with the most read component

pgurley:
https://jira.natgeo.com/browse/SYTO-1056

These changes address latest comments on SYTO-1056

To test:

- Pull SYTO-1056-3
- npm test
- gulp serve
- Margin when there's no image can be verified at http://127.0.0.1:3000/contentPackages/
- Padding around title when there is an image can be verified at http://127.0.0.1:3000/contentPackages/contentPackage/index.html


bpearsoncse:
Let's Go To Mexico :shipit:

PR #372, opened 2016-10-20

mwise-cse:
Fixed kicker alignment in large text only promo as per Kevin's comment.


bpearsoncse:
Tested locally works as advertised 👍

PR #382, opened 2016-10-25
Rename some files to unblock AEM finickyness

benjachapman:
AEM is björking when trying to compile a filename with two underscores (weird, beats me)


asabaylus:
Excellent!!!!

PR #390, opened 2016-10-27
[SYTO-1052] Parallaxing Rails do not always line up at bottom of scroll

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-1052


benjachapman:
LGTM

PR #392, opened 2016-10-28
[AEM-6604]: update Rails parallax logic to be sticky in some scenarios

benjachapman:
https://jira.natgeo.com/browse/AEM-6604

This work modifies the existing parallax behavior for the two instances of "Rails" (mostread/latest, hero with two rails)

When the shorter of the two rails is found to have a smaller height than that of the viewport, that element will now become sticky (fixed to the top of the page, with 40px of top margin) rather than parallaxed (that is, still moving, but at a different rate than the rest of the page). This will also apply when BOTH rails are shorter than the viewport, replacing the previous behavior in that situation. 

The logic is able to toggle between these modes on resize, should the smaller rail have become larger than the viewport or vice-versa


pgurley:
Test well. Merged

PR #393, opened 2016-10-28
[bug/SYTO-1119] - Video playlist sometimes does not change videos.

arieldelfor:
This PR intends to fix mortar videoplaylist when using latest modules-video (1.3.2).
You can test this only using the latest version modules video: blocker SYTO-1113
- git checkout bug/SYTO-1119
- gulp serve
- navigate to localhost:3000/components/
- scroll down to the videoplaylist, standing at there refresh the page.
- ensure the browser loads the page from that standing point and click on the videoplaylist items
- expected: the video should load.
---

JIRA: https://jira.natgeo.com/browse/SYTO-1119

### DIFF
--- a/src/modules/video/video.js
+++ b/src/modules/video/video.js
@@ -44,9 +44,13 @@ class Video extends Component {
     }
   }
 
+  componentWillUnmount() {
+    this.willUnmount = true;
+  }
+
   handleOnLoad() {
     // if we are not using editmode
-    if (this.props.isEditMode !== true) {
+    if ((this.props.isEditMode !== true) && (!this.willUnmount)) {
       this.createPlayer();
     }
   }
--- a/src/modules/video/video.js
+++ b/src/modules/video/video.js
@@ -12,6 +12,7 @@ class Video extends Component {
 
   constructor(props) {
     super(props);
+    this.handleOnLoad = this.handleOnLoad.bind(this);
     this.videoContainer = `<div id="${props.model.instance}" class="mt3_video-player" data-guid="${props.model.guid}"></div>`;
   }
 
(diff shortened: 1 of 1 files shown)


pmourelle:
Code looks good...

src/modules/video/video.js
```diff
@@ -68,24 +69,21 @@ class Video extends Component {
...
 
-    if (lazyLoad === true) {
-      return (
-        <figure itemType="http://schema.org/VideoObject" className={`mt3_video mt3_videopromo-container mt3_bgcolor--black ${className}` }>
-          <div className="mt3_video-wrapper" dangerouslySetInnerHTML={{__html: this.videoContainer}}>
-          </div>
-          <LazyLoad offsetVertical={200} onContentVisible={this.handleOnLoad.bind(this)}>
-            <span />
+    return (
+      <figure itemType="http://schema.org/VideoObject" className={`mt3_video mt3_videopromo-container mt3_bgcolor--black ${className}` }>
```
pmourelle: Notice that previous `else` had `mt3_bgcolor--gray5` as class for `<figure>`.

src/modules/video/video.js
```diff
@@ -68,24 +69,21 @@ class Video extends Component {
...
+        <div className="mt3_video-wrapper" dangerouslySetInnerHTML={{__html: this.videoContainer}}>
+        </div>
+        { lazyLoad &&
+          /** Don't use LazyLoad's onContentVisible callback.
+           * Lazyload triggers twice the callback if the element
+           * was on the viewport at first mount.
+           **/
+          <LazyLoad offsetVertical={200}>
+            <span ref={() => { this.handleOnLoad() }}/>
```
pmourelle: This `ref` doesn't look like the proper way of passing a callback. Previous way seems better for this line.

src/modules/video/video.js
arieldelfor: Yes, it is a fix for a bug never seen before, because we had never disabled lazyload.

src/modules/video/video.js
arieldelfor: What do you mean for previous way?
I left a comment in the code about why this hack is needed. I think the long term solution is to open a pull request at react-lazy-load repository, but i am afraid we don't have bandwidth to be awaiting 3rd party answers.

sebasrgl: nevermind, this only happen in my local env. Could anybody else test this?

krlospelaez:
LGTM :+1:

PR #396, opened 2016-10-28
[SYTO-1026] Removing space above hero when there is no heading

pgurley:
https://jira.natgeo.com/browse/SYTO-1026

Addresses use case for when there is a hero and no heading. This removes the spacing above the hero

To test:
Pull SYTO-1026

Open src/contentPackages/heroWithTwoRails/docs.html and remove
`       "heading":"RHUBARB PIE IS THE ORDER OF THE DAY",
`

Load http://localhost:3000/contentPackages/ and verify there is no space above the hero image


bpearsoncse:
Tested locally - looks as advertised 👍 :shipit:

PR #400, opened 2016-10-31
[AEM-6622] When only one image is used for a gallery card, none of the following cards render

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6622

**To Test in Mortar**:
- Delete the second image in leadMedia in promocard/docs.html
- verify card renders as an article 

**To Test in AEM**:
- Create a Gallery Page and leave the Gallery blank
- Add an image to the OG:Image field of the Gallery page (in page properties)
- Then use that page in the Page Path of the Story Card (leaving the override images blank)
- Verify card renders as an article


pgurley:
Merged

PR #401, opened 2016-10-31
[SYTO-1059] Fixed an issue that was causing some texts were displayed with Helvetica instead of Hass

krlospelaez:
To test: 

- `git checkout feature/SYTO-1059-3`
- Verify that text displayed using one of these classes (`mt3_h3, mt3_h5, mt3_sub1, mt3_sub2, mt3_sub3, mt3_sub4 and mt3_btn`) has font family haas.


arieldelfor:
LGTM

PR #402, opened 2016-11-01
[SYTO-1060] Adding large size heading

pgurley:
To test:
- Pull SYTO-1060
- gulp serve
- navigate over to http://192.168.1.108.xip.io:3000/contentPackages/heroWithTwoRails/
- Test large and standard sized heading options


benjachapman:
Looks great -tests well. Wonder if Kev will want to tweak font sizes, but that's none of my business

PR #406, opened 2016-11-02
[AEM-6648] Video Duration Not Displaying

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6648

To Test:
Pull AEM-6648
Verify duration shows both with and without the presence of a kicker

Also, I refactored the way the cards handle the z-index functionality of the kicker based on whether it is a link or not.


benjachapman:
LGTM - added some slight edits

PR #411, opened 2016-11-03
[SYTO-1128] Create Unit Test for EQ Component in Mortar

bpearsoncse:
https://jira.natgeo.com/browse/SYTO-1128


test/src/util/EQ_test.js
```diff
@@ -0,0 +1,104 @@
...
+    before(() => {
+
+      wrapper = mount(<EQ
+        sizeClasses={sizeClasses}
+      />);
+      wrapper.setState({ contentWidth: 420 });
+    });
+
+    it('Should have a div with small class', () => {
+      expect(wrapper.find('.small-class').type()).to.equal('div');
```
pmourelle: Class is actually called `small-class`?

benjachapman: that's just one of the mock classes being used in this test I reckon - this component accepts classes as arguments

bpearsoncse: ^Yeah, those are just test classes.

benjachapman:
LGTM

PR #413, opened 2016-11-04
SYTO-1155: modify HTR heading animation

benjachapman:
https://jira.natgeo.com/browse/SYTO-1155

This ticket is to have the Heading stop its parallax behavior 40px from the bottom of the Hero With Two Rails, rather than flush with the bottom

Also included:
- a correction that prevents large promos from having extra bottom-margin
- a tweak to the rate at which the parallax system rechecks the height of the document
- a tweak to the rate at which the HwTR heading will reach the center of the viewport

EDIT: updating this PR so that only includes the tweaks above - will add separate PR (https://github.com/natgeo/mortar/pull/414) for the 40px change, as I'm currently trying to talk Kevin into not doing it :)


pgurley:
These changes test well. Merged

PR #414, opened 2016-11-04
SYTO-1155 (part 2): 40px change for HwTR's heading

benjachapman:
https://jira.natgeo.com/browse/SYTO-1155

This change is so that the Heading will stop its parallax behavior 40px from the bottom of the Hero With Two Rails, rather than flush with the bottom


benjachapman:
not gonna do this after all

PR #417, opened 2016-11-07
[SYTO-1147] Disable opacity transition when using Chrome 54

krlospelaez:
This PR disabled opacity transition on Chrome 54 on versions less than 54.0.2840.87, the bug on Chrome causes a right rail flashing when you scroll down on the left rail and then hover/select other item in the right rail.


sebasrgl:
Close this PR since the approach used was implemented on https://github.com/natgeo/modules-global-nav/pull/353

PR #425, opened 2016-11-10
[AEM-6706] Sponsor Content Label text no longer appears in the POD Promo

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6706


pgurley:
Looks good to me. I updated the configurator to allow this to be tested more easily. Merged

PR #428, opened 2016-11-14
[SYTO-1167] Modify Form input Styling

hkekeli:
https://jira.natgeo.com/browse/SYTO-1167

Confirm the updated styling matches the mocks.


src/styles/patterns/_inputs.scss
```diff
@@ -32,5 +37,11 @@
 }
 
 .mt3_textarea {
+  @extend %mt3_haas;
+  font-weight: 400;
```
pmourelle: Looks like you're overriding here the styles of L7. Maybe you could move those styles with the default ones (L2-L5), unless there's some extra styling that mt3_verlag applies and you want to avoid (because mt3_haas doesn't override them).

src/styles/patterns/_inputs.scss
```diff
@@ -32,5 +37,11 @@
...
   min-height: $mt3_unit * 6;
+
+  &::placeholder {
+    @extend %mt3_verlag;
```
pmourelle: Placeholder should be inheriting already the fonts definitions. Is some special attribute to set?

src/styles/patterns/_radio.scss
```diff
@@ -14,7 +14,7 @@
       display: inline-block;
       height: $mt3_radio-height;
       margin-right: $mt3_unit * .75;
-      opacity: .9;
+      opacity: .6;
```
pmourelle: Better using a variable for this opacity, is being used at least twice.

src/styles/patterns/_inputs.scss
hkekeli: From I'm getting from the mocks and AC, the` textarea` uses two different fonts unlike the other input fields. The input fields are inheriting verlag directly, but the textarea  only uses Verlag on the `placeholder`.

src/styles/patterns/_inputs.scss
hkekeli: It's because of this line:
      `  <textarea class="mt3_input mt3_textarea ....></textarea>`

So on line 7 we're applying the font style to all `mt3_input` with the exception of  `mt3_textarea` since they use 2 different fonts. So we're not overriding any styles. 

```
&:not(.mt3_textarea) {
    @extend %mt3_verlag;
    font-weight: 400;
  }
```

src/styles/patterns/_radio.scss
hkekeli: will do!

src/styles/patterns/_inputs.scss
pmourelle: I see. Different font families for input/textarea is a discussion out of the scope of the ticket anyway. 👍

src/styles/themes/default/vars/_inputs.scss
```diff
@@ -6,6 +6,7 @@
 
 // Checkbox Input Values
 
+$mt3-input-default-opacity: .6;
```
pmourelle: 👍

src/styles/patterns/_inputs.scss
pmourelle: Same thing as last comment. It's good for me now. Font families to discuss later.

pmourelle:
LGTM!

PR #429, opened 2016-11-14
[AEM-6380] POD Promo Image Afilliation does not render

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6380


pgurley:
Tests well locally in Mortar. Merging

PR #436, opened 2016-11-16
[AEM-6765] Pipe overlaps POD Promo image

pgurley:
https://jira.natgeo.com/browse/AEM-6765

Removing transform on mobile where paint flashing isn't a problem, but the header being layered above the image is

To test:

- Run npm test
- Run gulp serve
- Navigate to http://localhost:3000/components/podpromo/
- At desktop verify hovering over the POD doesn't cause other parts of the page to repaint
- At mobile verify pipe attached to the header does not overlay image


bpearsoncse:
👍

PR #437, opened 2016-11-16
SYTO-1167 Modify Form Input styling

hkekeli:
https://jira.natgeo.com/browse/SYTO-1167

Updated font to bold and bumped the input font-size to 12px per Lauren's feedback.


pmourelle:
LGTM!

PR #438, opened 2016-11-17
[SYTO-1163] Mortar home page modifications

hkekeli:
https://jira.natgeo.com/browse/SYTO-1163

To test: 

- Check out the Brand new mortar homepage
- Make sure the AC is met


src/site/_styles/docs.scss
```diff
@@ -79,11 +71,6 @@ td, th {
   word-wrap: break-word;
 }
 
-.version {
```
pmourelle: This gives style to `version` div right below. Doesn't look too good, but is what we have so far :/

pmourelle: Nevermind, just saw how it looks without it. So cool!! :D

pmourelle:
It looks great! 👊

PR #442, opened 2016-11-21
[SYTO-1193-2] Text only cards should still get hover state on title regardless of breakpoint

pgurley:
https://jira.natgeo.com/browse/SYTO-1193

Retains hover state for text only cards when at the large breakpoint

To test:

- Pull SYTO-1193-2
- Run npm test
- gulp serve
- Navigate over to http://localhost:3000/components/
- Scroll down to the "Promo Card" section
- Verify titles get gray hover state at all breakpoints, and that cards with images only get a hover state when the title is not overlaying the image (large breakpoint)


benjachapman:
👍

PR #447, opened 2016-11-22
[SYTO-1230] Fix icon misalignment on icon page

hkekeli:
https://jira.natgeo.com/browse/SYTO-1230


pmourelle:
Seems that issue was some wording using 2 rows, min-height change helps to wrap this 2 lines. In case some icon takes more than 2 lines we'll be against the same issue again.
Refactoring this sections using flexbox or some other implementation would be a better approach probably. Could you create a ticket for it with very low priority?
PR is OK!

hkekeli:
@pmourelle  As I recall, this and the color page were previously written using flexbox but design wanted to use the specific widths and breakpoints provided by the grid. It's not perfect but it's a one line bugfix compared to the other bugs flexbox might come with :/

PR #451, opened 2016-11-28
[SYTO-899] Create secondary nav for components

hkekeli:
https://jira.natgeo.com/browse/SYTO-899

This update includes a pattern for the secondary navigation on Mortar.

On the component page, check out the secondary nav and make sure the links take you to the corresponding sections.


src/site/_layouts/default.html
```diff
@@ -175,6 +175,10 @@
             .removeClass('codesnippet-icon--expand');
        }
     });
+
+    $(function(){
+          window.mortarSite.renderSecondaryNavigation(document.getElementById('secondaryNav'));
```
pmourelle: Indentation :)

src/site/nav/SecondaryNav.js
```diff
@@ -0,0 +1,38 @@
...
+              <li key={anchor} className="secondarynav-list-item mt3_row-gut-half">
+                <a className="secondarynav-item-link"
+                  href={"#" + anchor}>
+                  {text}
+                </a>
+              </li>
+          )
+    });
+
+    return <div className="mt3_col-lg-3 secondarynav mt3_subh4">
```
pmourelle:
```
return (
  <div ...
)
```
would be more legible. Also there're a mix on 2 and 4 spaces of indent.

src/site/nav/SecondaryNav.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component }  from 'react';
+import * as ReactDOM from 'react-dom';
```
pmourelle: If `render` is only thing needed by ReactDom better to import just that.
`import { render as DOMrender } from 'react-dom';` for example?

pmourelle:
LGTM!

PR #454, opened 2016-11-29
[SYTO-1172] Use new heading classes on Mortar pages

hkekeli:
https://jira.natgeo.com/browse/SYTO-1172

Confirm the AC is met.


pmourelle:
Looks good.

PR #456, opened 2016-11-30
[SYTO-1219] Coverage into mortar

pmourelle:
Starting coverage with the bar medium low for not conflicting with any code/work already in progress. 
Some components needed to be refactored into class-components, due to conflicts with coverage and enzyme.
Some tests still need to be worked to fully test, in special the ones using `shallow()` as they aren't completing the full rendering of its children components (ie Video). A new ticket will be created for it.


sebasrgl:
PR OK!

PR #463, opened 2016-12-02
[AEM-6652] 4-Up margin in first 2 promos prevents even rails when scrolling

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6652

This may be tested in Mortar by deleting all cards except the first two, and in AEM (after copying /lib) any page that has a Content Package present such as http://aem.localhost.nationalgeographic.com:4502/content/ngdotcom/en_US/versions/v1.html,
which I believe should be present if the Homepage content package was installed previously.


pgurley:
Looks good

PR #469, opened 2016-12-05
Reorder Content Packages

hkekeli:
https://jira.natgeo.com/browse/SYTO-1181

Reordered content packages like so: 
-Broadsheet
-Content Package
-Hero with Two Rails
-POD Promo
-Video Playlist


pmourelle:
LGTM

PR #470, opened 2016-12-05
[AEM-6652v2] 4-Up margin in first 2 promos prevents even rails when scrolling

bpearsoncse:
https://jira.natgeo.com/browse/AEM-6652

This is to fix a bug found in testing, where if the four-up only has the first card it does not align.


pgurley:
Merged

PR #480, opened 2016-12-07
[SYTO-1175] Update background color on inverse

hkekeli:
https://jira.natgeo.com/browse/SYTO-1175

Confirm the AC is met.


krlospelaez:
LGTM :+1:

PR #482, opened 2016-12-08
[SYTO-1266] Remove some components from Mortar

hkekeli:
https://jira.natgeo.com/browse/SYTO-1266

Removed some components from Mortar. Check AC for the list.


krlospelaez:
LGTM :+1:

PR #491, opened 2016-12-12
[SYTO-1282] Videos in the video playlist on Homepage do not always change when playing

krlospelaez:
This PR fixes next bugs on video playlist:

- When breakpoint less than 70em was causing videos won't load.
- Video player won't play videos when description has the same size as the last reproduced video.


benjachapman:
LGTM, tests well

PR #503, opened 2016-12-14
[AEM-6859] Fixing display with firefox and append tracking codes to heroUrl

pgurley:
https://jira.natgeo.com/browse/AEM-6859

Fixes a display issue with Firefox and the CTA button.
Also appends trackingCodes to the heroUrl

Testing is the same as for the other broadsheet stories


benjachapman:
👍

PR #509, opened 2016-12-15
[SYTO-901] Add secondary nav to components page

hkekeli:
https://jira.natgeo.com/browse/SYTO-901

To Test, 

Go to components page 
Verify it contains a left rail menu and that all links work.

Bonus:

- Fixed ESC key bug on site nav


src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
...
-    const secondaryNavClass = cx(
-      'mt3_col-lg-3',
-      'secondarynav',
-      'mt3_subh4', {
-      'secondarynav-fixed': this.state.fixed
-    });
+  render() {
+    const fixedClass = this.props.sticky
+      ? cx({
+        'secondarynav-fixed': this.state.scrolledPastStart && !(this.state.closeToBottom && this.state.navIsLong),
```
pmourelle: You may want to have this in a `const` before the `cx()` to simplify it and not calculating again.

src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
...
+window.mortarSite.renderSecondaryNavigation = function (el) {
   if (!el) {
     return;
   }
-
   if (el.classList.contains('secondary-nav-sticky')) {
-    ReactDOM.render(<SecondaryNav sticky />, el);
-  } else {
+    ReactDOM.render(<SecondaryNav sticky/>, el);
+  }
```
krlospelaez: Use `} else {`

pmourelle:
LGTM!

src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
...
+  handleScrolling() {
+    const body = document.body,
+          html = document.documentElement;
+    let documentHeight = Math.max(body.scrollHeight, body.offsetHeight,
+        html.clientHeight, html.scrollHeight, html.offsetHeight);
+    let scrolledPastStart = window.pageYOffset > this.menuStartPosition;
+    let closeToBottom = window.pageYOffset + window.innerHeight + MIN_DISTANCE_FROM_BOTTOM > documentHeight;
+
+    this.setState({
+      documentHeight: documentHeight,
```
krlospelaez: `documentHeight`, `pageY`, `startPos` and `ctb_debug` is not in use

src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
...
+      pageY: window.pageYOffset,
+      startPos: this.menuStartPosition,
+      ctb_debug: window.pageYOffset + window.innerHeight + MIN_DISTANCE_FROM_BOTTOM,
+      scrolledPastStart: scrolledPastStart,
```
krlospelaez: Don't need to specify scrolledPastStart again. Writing `scrolledPastStart,` is enough

src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
...
+      closeToBottom: closeToBottom,
```
krlospelaez: Same here

src/site/nav/SecondaryNav.js
```diff
@@ -1,94 +1,118 @@
 import cx from 'classnames';
-import React from 'react';
+import React, { Component }  from 'react';
 import ReactDOM from 'react-dom';
 import {throttle, zip } from 'lodash';
 
-const SECONDARY_NAV_FIXED_TOP = 90; // px
```
krlospelaez: You still need this. It's required to prevent the menu goes behind the header.

hkekeli: This is no longer an issue

hkekeli:
Updated

krlospelaez:
LGTM :+1:

src/site/nav/SecondaryNav.js
```diff
@@ -1,93 +1,113 @@
...
+          html = document.documentElement;
+    let documentHeight = Math.max(body.scrollHeight, body.offsetHeight,
+        html.clientHeight, html.scrollHeight, html.offsetHeight);
+    let scrolledPastStart = window.pageYOffset + SECONDARY_NAV_FIXED_TOP > this.menuStartPosition;
+    let closeToBottom = window.pageYOffset + window.innerHeight + MIN_DISTANCE_FROM_BOTTOM > documentHeight;
+
+    this.setState({
+      scrolledPastStart,
+      closeToBottom,
+    });
```
benjachapman: This will call setState every time the function is called - it should only call setState when one of those variables has changed, so it doesn't re-render unnecessarily

pmourelle: Maybe for having in `this.position` for example, instead of state?

hkekeli: @benjachapman @pmourelle 

 To follow the React ethos, it would be better to keep things that directly affect the render in State.

If we’re worried about re-rendering performance, we can use `shallowCompare`. I updated [SYTO-901-b](https://github.com/natgeo/mortar/pull/524) to show this.

The problems with not using a state are:

1. We do need it to re-render *sometimes*, and we won’t get that unless we use state.
2. **Unit testing**— (if/when we get to it) will be easier when the state/props convention is followed.
3. **Debugging is simplified** because we can check on the current state of the item via React dev tools, and change it directly via the same dev tools. 
4. **Potential bugs**: React doesn’t actually guarantee that it’ll keep anything alive except state/props in its internal tree. Our components might break in later versions of React.
5. **Hot reloading**: It’s based on the states/props convention. If we use internal variables, we lose the possibility to hot-reload our components in the future. 
6. **Philosophical issues**: state is our single source of truth.

PR #511, opened 2016-12-15
[SYTO-1267] Fixed an issue on CTA button when type is link

krlospelaez:
To test:

- Go to Components
- Search for the link "share this article"
- Verify it has black font color


src/modules/button/CTAButton.js
```diff
@@ -49,7 +49,8 @@ class CTAButton extends Component {
         'mt3_fullwidth': true,
         'mt3_btn--naked': true,
         'mt3_btn--naked--reversed' : inverse,
-        'mt3_btn--naked--inactive': inactive
+        'mt3_btn--naked--inactive': inactive,
+        'mt3_cta-link': true
```
pmourelle: All the `true`s could be put as plain strings before the `{`.

{{{
classnames(
  'mt3_btn',
  ..., {
    'mt3_btn--naked--reversed' : inverse,
    ...
  })

src/modules/button/_styles.scss
```diff
@@ -16,6 +16,10 @@
   }
 }
 
+.mt3_cta-link {
```
pmourelle: "cta" standd for...?

krlospelaez: CTA = Call to Action. It's the name for this component. https://github.com/natgeo/mortar/blob/master/src/modules/button/CTAButton.js

src/modules/button/CTAButton.js
pgurley: @hkekeli It's using https://www.npmjs.com/package/classnames to conditionally add classnames based on props and such

src/modules/button/_styles.scss
pmourelle: Dah. 👍 Sorry Carlos, got a brain fart.

src/modules/button/CTAButton.js
pmourelle: It's called "Patefuá", but is secret, don't tell anyone.
Is just setting explicitly the list of always-true-classes because there's no need to calculate anything, appart from the calculated-classes.

krlospelaez:
Changes done

src/modules/button/CTAButton.js
```diff
@@ -44,13 +44,17 @@ class CTAButton extends Component {
...
-        'mt3_btn--naked--inactive': inactive
-      });
+      classes = classNames(
+        'mt3_btn',
+        'mt3_fullwidth',
+        'mt3_fullwidth',
+        'mt3_btn--naked',
+        'mt3_cta-link',
+        {
+          'mt3_btn--naked--reversed' : inverse,
```
pmourelle: Last but not least, the space before `:`. I'm pretty sure you have a brand new sass-lint warning about it.

hkekeli:
LGTM

PR #520, opened 2016-12-16
add jQ and dotdotdot to VideoThumbnail

benjachapman:
to replicate the error in Mortar, have to comment out this line, so that a global $ object does not exist: https://github.com/natgeo/mortar/blob/master/src/site/_layouts/default.html#L20

to test: confirm that commenting out the above ^ + this ticket's work allows VP to play videos


pgurley:
Looks good

PR #524, opened 2016-12-19
Code updates

hkekeli:


src/site/nav/SecondaryNav.js
```diff
@@ -36,6 +37,10 @@ export default class SecondaryNav extends React.Component {
     window.removeEventListener('scroll', this.handleScrolling);
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
```
pmourelle: A couple of things:
1) Take note 1st line of https://facebook.github.io/react/docs/shallow-compare.html. PureComponent should be the one to use as it has already implemented the `shouldComponentUpdate`.
2) The shallow comparing won't change a thing as the state is still changing on each scroll. If state changes render happens anyway.

Due to 2) this is a no-go solution. IMO the better way would be doing the `cx()` checks *before* doing the `setState()`, so state will hold such values and render will be simpler. This would need also the `PureComponent` instead of `Component`.

pmourelle: Point 1) is valid in "some" recent version of react. We're giving support also to v0.14 so `shallowCompare` seems ok.
Point 2) would be already taken care of. My bad, I thought the values saved were `px` or volatil value instead of final boolean compare.

